PR #1, opened 2015-06-11
Use Apt module version 2.x

runephilosof:


runephilosof:
I realized that I am not using Apt 2.x myself...
So I haven't tested this.

Besides there is a syntax error because of missing ','

tobru:
Thanks for your PR. I consider adding support for the apt 2.x version in the future.

PR #2, opened 2015-06-16
RHEL 7 support

danfoster:
This PR addresses some problems to fix some issues RHEL 7 support.

1. Use `$releasever` instead of the hardcoded `5` in the `baseurl` for the `yumrepo`.
2. Don't force the `upstart` service type if it's a redhat family machine. Let puppet infer that it's `systemd`.
3. Because the chef installer symlinks the systemd service to `/etc/systemd/system/default.target.wants/gitlab-runsvdir.service`, it is not enableable. But it gets started on boot due to it being under `default.target.wants`. The work-around is just to tell puppet not to try and enable the service.

For info, you get the following error in you try to enable the service.
```
# systemctl enable gitlab-runsvdir
Failed to issue method call: No such file or directory
```

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -30,7 +30,7 @@ class gitlab::install {
       'redhat': {
         yumrepo { 'gitlab_official':
           descr         => 'Official repository for Gitlab',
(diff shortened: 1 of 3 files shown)


tobru:
Thanks a lot for this PR

PR #5, opened 2015-06-22
Remove rails; it's a duplicate of gitlab_rails.

### LINKED ISSUE #3: rails vs gitlab_rails?
You seem to have two parameters, rails and gitlab_rails. I expected to use gitlab_rails, but looking at the template it uses rails. This is confusing. I expect rails should be deprecated in favour of gitlab_rails.

tdb:
I chose gitlab_rails because it matches the name of the configuration section and because it's what's used in README.md.

This resolves #3.


tobru:
Thanks for this PR, I'll look into it today and merge it then...

PR #7, opened 2015-07-06
Omnibus service fix and template improvement

jrwesolo:
I was testing this on CentOS 6.6 and was running into issues with the service being managed correctly. Since we are using the omnibus package regardless of os, let's use the gitlab-ctl commands all the time. I also experienced an issue where the gitlab.rb template was changing the order of config lines on subsequent runs which would result in unnecessary restarts of gitlab. I changed the template to sort the keys of the hash first and that alleviated the problem. This is nice in general for a clean and organized config.


tobru:
Thanks a lot for this great contribution. Will do a new release today...

PR #10, opened 2015-07-10
Fixed bug with gitlab.rb.erb template for use with git_data_dir

sd-robbruce:


tobru:
Thanks for finding and fixing!

sd-robbruce:
You're welcome, let me know when you release a new version on puppet forge

PR #12, opened 2015-07-10
Fixed issue with using $releasever for RedHat

sd-robbruce:
If using RedHat 6.5, using $releasever in the yum repository fails with the following message

```
Error: Execution of '/usr/bin/yum -d 0 -e 0 -y list gitlab-ce' returned 1: Error: Cannot retrieve repository metadata (repomd.xml) for repository: gitlab_official. Please verify its path and try again
```

Using the `$::os[release][major]` fact fixes this issue, the stringify_facts option must be set to false for this to work.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -29,9 +29,11 @@ class gitlab::install {
         }
       }
       'redhat': {
+        $releasever = $::os[release][major]
+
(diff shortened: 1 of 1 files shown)


sd-robbruce:
Closing as moving to separate branch

PR #14, opened 2015-07-13
RedHat releasever

sd-robbruce:
If using RedHat 6.5, using $releasever in the yum repository fails with the following message

```
Error: Execution of '/usr/bin/yum -d 0 -e 0 -y list gitlab-ce' returned 1: Error: Cannot retrieve repository metadata (repomd.xml) for repository: gitlab_official. Please verify its path and try again
```

Using the `$::os[release][major]` fact fixes this issue, the stringify_facts option must be set to false for this to work.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -30,9 +30,11 @@ class gitlab::install {
         }
       }
       'redhat': {
+        $releasever = $::os[release][major]
+
         yumrepo { 'gitlab_official':
           descr         => 'Official repository for Gitlab',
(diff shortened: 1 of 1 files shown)


jantman:
Can you please add a note somewhere that ``stringify_facts`` must be set to false for this to work? It's still failing in my environment because of that...

tobru:
@jantman done in https://github.com/vshn/puppet-gitlab/commit/482f17e2ab0f8ef77dac5580009b58789e24d5e4

PR #15, opened 2015-07-13
Added check for external database

sd-robbruce:
If using an external database, an additional exec step is needed as outlined in the gitlab documentation:

https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/doc/settings/database.md

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -53,6 +53,15 @@ class gitlab::config {
       refreshonly => true,
       timeout     => 1800,
     }
+
+    unless $postgresql[enable] {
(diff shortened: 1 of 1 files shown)


tobru:
Thanks a lot for your PR. I'll wait with merging until you give me an OK, is that fine for you?

sd-robbruce:
Yeah sure, just an edge case I found as I'm using an external db on AWS for HA.

tobru:
Every PR is welcome =)

sd-robbruce:
Confirmed that this works in a HA scenario, when coupled with the high_availability option.

PR #24, opened 2015-08-12
better support for older puppet versions, see PUP-1597

### LINKED ISSUE #20: Question: Would this module work with puppet 3.6.2?
We were running spuder's gitlab module, and really want to leap to this module, but we are running puppet enterprise 3.6.2. Upgrading is on the roadmap, but not in the immediate future. Is it possible that the module is compatible with 3.6.2, but untested? Or did you use some 3.7-specific functionality (which is what I'm guessing).

spielkind:
This commit fixes #20

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -247,8 +247,8 @@ class gitlab (
   class { '::gitlab::config': } ~>
   class { '::gitlab::service': }
 
-  contain ::gitlab::install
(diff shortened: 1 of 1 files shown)


tobru:
nice catch! thank you

tobru:
I had this already locally, sorry. Thanks for your PR, it's now integrated.

PR #29, opened 2015-09-29
Fix unrecognized datatypes inside array in decorate method

deadratfink:


deadratfink:
Hi,

first thx for the nice module. I think I found an error when handling param values in the gitlab.rb.erb.
The current impl does not allow to process e.g. hashes which are handled in the decorate(v) method, for instance:

```
      gitlab_rails => {
...
          'omniauth_providers'                => [
              {
                  "name"  => "crowd",
                  "args"  => {
                      "crowd_server_url"     => "my_url",
                      "application_name"     => "Gitlab",
                      "application_password" => "my_passwd"
                  }
              }
          ],
...
      },
```
will result in:

```
gitlab_rails['omniauth_providers'] = ["{"name"=>"crowd", "args"=>{"crowd_server_url"=>"my_url", "application_name"=>"Gitlab", "application_password"=>"my_passwd"}}"]
```

which is incorrect and leads to an error in provision, it should be without the double colons around the hash:

```
gitlab_rails['omniauth_providers'] = [{"name"=>"crowd", "args"=>{"crowd_server_url"=>"my_url", "application_name"=>"Gitlab", "application_password"=>"my_passwd"}}]
```

Best Regards
Jens

tobru:
@deadratfink Thanks a lot for your PR. This is indeed an error inside the decorate function. Before I can merge your fix I would appreciate if you could fix the tests accordingly.

deadratfink:
Adapted test now, pls check, sorry for inconvenience :-)

tobru:
@deadratfink thanks for fixing the test. looks good to me now.

PR #34, opened 2015-10-09
allow disabling of omnibus user management

alexsmithhp:
add ability to change manage_accounts hash to disable omnibus account management
tested on centos 7


tobru:
thanks for the PR

PR #40, opened 2015-11-20
Add 'manage_package' parameter

MrStaticVoid:
Some installations require modification before configuration, thus it
is desirable to be able to unmanage the package in this module, install
and modify it with another, then use this module to configure the
installation.  As an example, it is nice to be able to add additional
omniauth providers after installing the omnibus package, but before
configuring it.


tobru:
Thanks for your contribution!

PR #41, opened 2015-11-30
Fix this module should work with out having to set the stringify_facts option

jcsmith:
Since falling back to $releasever will likely never work as it espands to 7server or 6server everywhere I tested update the module to use the $::operatingsystemmajrelease fact when the is_hash test fails for $::os.  There may be a better way to handle this and I am willing to change it with proper feedback.  I would also like to understand the reasoning behind usng the $:os hash instead of the $::operatingsystemmajrelease fact. Maybe the best way would be to just use this instead?

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -52,7 +52,7 @@ class gitlab::install {
         if is_hash($::os) {
           $releasever = $::os[release][major]
         } else {
-          $releasever = "\$releasever"
+          $releasever = $::operatingsystemmajrelease 
         }
 
         yumrepo { 'gitlab_official':


jcsmith:
Just curious - are there plans to pull in this fix or something similar to address the issue?  I'm more than happy to implement/test a different fix if you provide some details in the direction you are heading.

tobru:
It's a bit hard for me to judge this change because I don't have a CentOS/RHEL to test it. But I probably thing this is a save one. So I'll merge it now. There will be a release in the next days...

jcsmith:
Great, thanks, if you need access to a RHEL box for testing this module on I could probably make that happen if you are interested. 

--
Josh Smith
KD8HRX

Email/jabber: juicewvu@gmail.com
Phone: 304.237.9369(c)

tobru:
Your contribution is in `1.7.1`. Enjoy :smile:

jcsmith:
Thank you - have a great new year!

Issue #27, opened 2015-09-02
gitlab-runsvdir.service not enabled running on CentOS 7


gwarf:
Hi,
Using `vshn/gitlab v1.5.0` on `CentOS 7` with `gitlab-ce 7.14.1` omnibus package it seems that the `gitlab-runsvdir` systemd service is not running, and `/usr/bin/gitlab-ctl` seems not able to start the gitlab services without having it running. As `gitlab-runsvdir` systemd service is not enabled gitlab also does not start automatically on boot.

``` sh
root@sources ~ # /usr/bin/gitlab-ctl status
fail: ci-redis: runsv not running
fail: ci-sidekiq: runsv not running
fail: ci-unicorn: runsv not running
fail: logrotate: runsv not running
fail: nginx: runsv not running
fail: postgresql: runsv not running
fail: redis: runsv not running
fail: sidekiq: runsv not running
fail: unicorn: runsv not running
root@sources ~ # /usr/bin/gitlab-ctl start
fail: ci-redis: runsv not running
fail: ci-sidekiq: runsv not running
fail: ci-unicorn: runsv not running
fail: logrotate: runsv not running
fail: nginx: runsv not running
fail: postgresql: runsv not running
fail: redis: runsv not running
fail: sidekiq: runsv not running
fail: unicorn: runsv not running
root@sources ~ # systemctl status gitlab-runsvdir.service                                                                                                      
gitlab-runsvdir.service - GitLab Runit supervision process
   Loaded: loaded (/usr/lib/systemd/system/gitlab-runsvdir.service; disabled)
   Active: inactive (dead)

root@sources ~ # systemctl enable gitlab-runsvdir.service                                                                                                      
ln -s '/usr/lib/systemd/system/gitlab-runsvdir.service' '/etc/systemd/system/basic.target.wants/gitlab-runsvdir.service'
root@sources ~ # /usr/bin/gitlab-ctl start                                                                                                                     
fail: ci-redis: runsv not running
fail: ci-sidekiq: runsv not running
fail: ci-unicorn: runsv not running
fail: logrotate: runsv not running
fail: nginx: runsv not running
fail: postgresql: runsv not running
fail: redis: runsv not running
fail: sidekiq: runsv not running
fail: unicorn: runsv not running
root@sources ~ # systemctl start gitlab-runsvdir.service                                                                                                       
root@sources ~ # /usr/bin/gitlab-ctl status                                                                                                                    
[...]

PR #43, opened 2015-12-17
service enabled for all distro's #27

witjoh:
As the documentation says, that the default for service_enabled is true, this should be true then for all distros.  
This is tested on centos7.1 and the latest gitlab 8.2.3

Issue #27 could be closed with this PR.

No rspec is changed since this kind of test is not covered.


tobru:
Yes, this makes sense. Thanks for the PR

PR #44, opened 2015-12-21
Added code for configuration of a cli and adding of a user with that cli

walterheck:
With standard gitlab omnibus installer for gitlab 8 there's no cli to interface with gitlab. This PR configures the gitlab gem and a config file that then allows the gitlab::user class to create users.

Eventually the gitlab::user class needs to become a custom type and provider, but for now this will do.


tobru:
Yay, this is a very nice addition, thank you very much. But before I can merge it I'd like to ask you for some documentation: In the .pp's and also in the README. And it would be great if you could fix some styling and alignement issues. This module tries to be conform to the official Puppetlabs style guide.
Thanks a lot for your contributiion!

tobru:
Oh and I think there is some code missing: You refer to parameters in params.pp which don't exist and are also not configurable in init.pp. I think it would be better to reference to parameters in init.pp so that they can be savely manage from hiera or other classes.

tobru:
And if you're able to add some Rspec tests, that would be fantastic!

walterheck:
Right after this I actually decided to bite the bullet and start work on a proper gitlab_user resource type. I guess I'll work on that instead and replace this PR when it's done.

tobru:
Sounds great... Looking forward to your type!

walterheck:
@tobru currently struggling with https://github.com/NARKOZ/gitlab/issues/159, frustrating!

tobru:
Wait for a new implementation of this PR

PR #52, opened 2016-01-20
As this template is writing a config file based on some hashes and ha…

msutter:
…shes are not meant to be in a certain order,

We need to ensure that the order of the keys are always outputed in the same order.
Otherwise puppet will always update(change) the config file.


tobru:
Great contribution, thanks a lot. Would you mind having a look at the failed checks? As this modules should be compatible with Puppet 4, it should also pass the future parser tests.

msutter:
Hi,

The tests are ok now.

Cheers
Marc

tobru:
great! thanks

Issue #66, opened 2016-03-14
gitlab::gitlab_rails hash merge not working


cdenneen:
I believe the issue is how this is being read in vs actually merging the data because hiera works fine.

cdenneen:
Here is the fix:
```puppet
  $gitlab_rails = hiera_hash('gitlab::gitlab_rails',$::gitlab::gitlab_rails)
```

cdenneen:
Reason is explained here:
https://docs.puppetlabs.com/hiera/3.0/puppet.html#limitations

tobru:
As a workaround you could create a "wrapper" module (f.e. profile_gitlab) which collects the needed information from Hiera using hiera_array() and pass it to the gitlab module...

PR #67, opened 2016-03-15
Fixes issue for gitlab_rails merging

cdenneen:
Automatic parameter lookup can't merge values from multiple hierarchy levels (https://docs.puppetlabs.com/hiera/3.0/puppet.html#limitations)

This is to fix vshn/puppet-gitlab#66


tobru:
We would prefer to have a "clean" solution which applies to all parameters with the data type Array and also prefer to have it in init.pp

PR #69, opened 2016-03-29
Allow settings of custom gitlab.rb config file

agray1017:
Gives the ability to set the gitlab.rb config file to your own puppet path overriding hiera.


tobru:
Thanks for your PR, good idea. Would you mind checking the failing tests and update them accordingly to your change?

PR #77, opened 2016-06-01
Add registry configuration

bashou:
It's a first step to configure new registry features with Gitlab 8.8.x


bashou:
Sh*t, added 2 days ago 👍

PR #76, opened 2016-05-27
Add support for Registry

llauren:
This adds support to the registry features introduced in GitLab 8.8, in response to issue #74. This is also my first ever contribution to an open source project, so please be gentle. I couldn't find any directives to how to actually accomplish this so i asked around and tried my best... 😕

Issue #74, opened 2016-05-25
Add support for Docker registry


tobru:
Add support for the new feature: Docker registry. See http://docs.gitlab.com/ce/administration/container_registry.html#container-registry-domain-configuration

llauren:
I've created a patch for this, which i'll try to pull-request. I couldn't find the directives on how to do this, so i just tried my best. This'll be my first ever contribution to an open source project, so please be gentle :)

tobru:
Thanks for the PR, merged. After testing it a new release will be made.

PR #84, opened 2016-07-02
Added registry configuration hash option

jkroepke:
Relates to #74, #76 
Credits to #77 

Added missing registry configuration hash. See https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template#L319


tobru:
Thanks a lot!

hypnotortoise:
Can you create a new forge release based on #84 please? This is much needed for people with multiple appserver behind external reverse proxy servers (e.g. change `token_realm` to HTTPS instead of HTTP, `registry_http_addr` for allowing none-localhost connections from proxys..)

tobru:
Will do a release asap

PR #90, opened 2016-08-22
Feature/master/systemd compatibility

bjeske:
Ensure => 'absent' for initd softlink on systems with systemd like ubuntu 16.04 etc.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -28,6 +28,12 @@
 #   Default: true
 #   Run the system service on boot.
 #
+# [*service_initd_ensure*]
+#   Default for 'Ubuntu 15.04, 15.10, 16.04, 16.10' or 'Debian 8': "absent"
+#   Else: "link"
+#   Sets "ensure => 'absent'" or "ensure => 'link'" on init.d softlink
+#   depending on systems OS to avoid conflicts with systemd.
+#
 # [*service_exec*]
 #   Default: '/usr/bin/gitlab-ctl'
 #   The service executable path.
@@ -253,6 +259,7 @@ class gitlab (
   $package_pin = $::gitlab::params::package_pin,
   # system service configuration
   $service_enable = $::gitlab::params::service_enable,
+  $service_initd_ensure = $::gitlab::params::service_initd_ensure,
   $service_ensure = $::gitlab::params::service_ensure,
   $service_group = $::gitlab::params::service_group,
   $service_hasrestart = $::gitlab::params::service_hasrestart,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -29,10 +29,10 @@
 #   Run the system service on boot.
 #
 # [*service_initd_ensure*]
-#   Default for 'Ubuntu 15.04, 15.10, 16.04, 16.10' or 'Debian 8': "absent"
+#   Default for systemd systems, as determined by the $::gitlab_systemd fact: "absent"
 #   Else: "link"
 #   Sets "ensure => 'absent'" or "ensure => 'link'" on init.d softlink
-#   depending on systems OS to avoid conflicts with systemd.
+#   depending on the $::gitlab_systemd fact to avoid conflicts with systemd.
 #
 # [*service_exec*]
 #   Default: '/usr/bin/gitlab-ctl'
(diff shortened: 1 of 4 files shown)


manifests/params.pp
```diff
@@ -31,6 +31,12 @@
     $service_enable = true
   }
 
+  if ($::operatingsystem == 'Ubuntu' and $::operatingsystemrelease in ['15.04','15.10','16.04','16.10']) or ($::operatingsystem == 'Debian' and $::operatingsystemrelease == '8') {
```
petems: It might be better to simplify this to use the systemd fact:

https://github.com/camptocamp/puppet-systemd/blob/master/lib/facter/systemd.rb

This would mean this patch would also work on RHEL flavoured systems with systemd (Centos 7, Oracle Linux 7, Fedora 14+ etc)

bjeske: This is not a core fact, right? So we would need a dependency for puppet-systemd... I'm not so sure if this is a good decision? As of Ubuntu 16.04 with facter 1.7.5 this fact is not present.

petems: It's not a core fact unfortunately no. Adding a dependency is one option, but seems a bit heavy handed for just one fact. 

However, it could be added as a fact to this module with it's own namespace (eg. gitlab_systemd_check or something), using only the code for the systemd check from the linked fact:

```ruby
Facter.add(:gilab_systemd_check) do
  confine :kernel => :linux
  setcode do
    Facter::Core::Execution.exec('ps -p 1 -o comm=') == 'systemd'
  end
end
```
The benefit of adding the fact would be that the code becomes a lot more manageable:

```puppet
if ($::gilab_systemd_check) {
```

bjeske: It seems "Facter::Core::Execution" is not present in facter 1.7.5, which comes with Ubuntu 16.04.
I had to use the "old %x" Sysntax to make this work:

```ruby
Facter.add(:gitlab_systemd) do
  confine :kernel => :linux
  setcode do
    %x{ps -p 1 -o comm=}.gsub("\n",'') == 'systemd'
  end
end
```
Or did i miss something here?

petems: Using `backticks/%x` or `Kernel.exec` will ignore the confine, so it will run on any system rather than be restricted to Linux (eg. Windows, OSX, BSD).

Good spot, It should be `Facter::Util::Resolution.exec` which is backwards compatible:
http://www.rubydoc.info/gems/facter/1.7.5/Facter/Util/Resolution.exec
http://www.rubydoc.info/gems/facter/2.4.6/Facter/Util/Resolution.exec

```ruby
Facter.add(:gitlab_systemd) do
  confine :kernel => :linux
  setcode do
     Facter::Util::Resolution.exec('ps -p 1 -o comm=') == 'systemd'
  end
```
end

I'll open a PR against the systemd repo for that to fix that! 👍

manifests/init.pp
```diff
@@ -28,6 +28,12 @@
 #   Default: true
 #   Run the system service on boot.
 #
+# [*service_initd_ensure*]
+#   Default for 'Ubuntu 15.04, 15.10, 16.04, 16.10' or 'Debian 8': "absent"
```
petems: This probably needs changing to "default for systemd systems, as determined by the $:: gitlab_systemd fact

manifests/params.pp
bjeske: Seems like i was a bit fast claiming 16.04 comes with 1.7.5... Not so sure if we enforce this facter version for one reason or another somewhere... 2.4.6 seems to be in 16.04.

petems: Heh, I think it depends if you're using Ubuntu upstreams or Puppet's ones 😄 

Either way, Util resolution is the right way 👍

tobru:
@bjeske Thanks a lot for your contribution. Before I can merge it could you please make sure that all tests passes? And if you could do a rebase so we only have one commit to merge would also be great.

petems:
@tobru I've rebased, squashed and added tests for this PR in #94, as I'm looking to move to the upstream for a customer rather than a fork. 

@bjeske The commit is still under your name in the new PR, but squashed and tests fixed

tobru:
Merged in #94

PR #91, opened 2016-08-23
Allow "Disable storage directories management"

gdowmont:
In environments that use shared  storage (ex. NFS) required directories might/will already exist. Running puppet module against a setup like that will try to re-create the directories and change their permissions. This stops GitLab from functioning correctly as it is not able to access the filesystem. This commit fixes it by allowing the users to set 'manage_storage_directores[enable]=false' as per https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/doc/settings/configuration.md#disable-storage-directories-management


tobru:
Thanks a lot for your contribution

PR #93, opened 2016-09-10
Added external_port parameter

blakejakopovic:
I'm using Gitlab that sit's behind an Nginx reverse proxy, and needed to change the external port. So I've added the `external_port` parameter. Note that to successfully use a different port, you need to add the port to the `external_url` as well. I'm currently using this fork, and it works as expected. I've also added a basic test. Please let me know if you have any concerns.

```puppet
class { 'gitlab':
  external_url  => 'http://domain.name:5555',
  external_port => 5555,
}
```


tobru:
Great! Thanks a lot for your contribution.

PR #94, opened 2016-09-16
Add systemd_compatibility

petems:
* Ensure absent softlink in init.d for operating systems with systemd.
* Updated init.pp description of "service_initd_ensure" to reflect recent changes.
* Added custom fact gitlab_systemd (duplicates fact from camptocamp-systemd, but we dont want full module for just one fact)
* Adds spec to test new systemd functionality
* New fact added to all tests to avoid errors with `STRICT_VARIABLES=yes`

Rebase, squash with tests from #90

### DIFF
--- a/lib/facter/gitlab_systemd.rb
+++ b/lib/facter/gitlab_systemd.rb
@@ -1,0 +1,18 @@
+# Fact: gitlab_systemd
+#
+# Purpose: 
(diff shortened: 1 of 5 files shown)


tobru:
@petems and @bjeske Thanks a lot for your effort! Will create a new module release soon.

PR #95, opened 2016-09-19
Fixes beaker tests

petems:
* Beaker tests when run in a small enough container (such as a Travis box) the Gitlab app can fail with (`Whoops, GitLab is taking too much time to respond.`)
* So we change the regex to see that also, so we can at least see the app has been attempted to start


tobru:
Thanks. But tests are still failing =(

PR #96, opened 2016-09-20
Refactor beaker tests for Travis

petems:
* There are issues running docker in docker in Travis, so lets test the easiest use-case
* Adds a specific Travis spec with the docker workarounds required
* Revert the class spec so tests in Vagrant can still work outside of Travis


tobru:
@petems thanks for the update. Would you mind adding docs for the new parameter before I merge it?

petems:
@tobru Done! 👍

tobru:
Great, thanks a lot. Let's see what Travis says...

PR #102, opened 2016-12-12
Fix #82

### LINKED ISSUE #82: Upgrade from CE to EE doesn't work without manual work
At least on RHEL, as you use a single `yumrepo` resource with the same name for 2 different repos, when upgrading from CE to EE edition the Yum repo is not being updated and upgrade does not happen until you manually remove the old repo and do a `yum clean`.

I think it would be enough to add edition to `yumrepo` resource name but unfortunately I don't have time to create a PR right now, sorry.

dhollinger:
This PR should fix the issue where upgrading from Gitlab CE to Gitlab EE require manual work.

Please review, may need to add additional code to remove the Repo for whichever edition is not being used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ spec/fixtures
 
 log/
 junit/
+
+.ruby-version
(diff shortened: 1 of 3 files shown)


dhollinger:
I did not add explicit tests for gitlab_official_ee repo files. I can do that as well if desired.

tobru:
Looks good so far. Merging it now and will give it a try.

PR #105, opened 2016-12-29
cirunner: add missing hard dependency for xz-utils

roock:
cirunner requires xz-utils package for docker to work correctly


tobru:
Great! Thanks a lot

PR #107, opened 2017-01-10
removing unneeded fact evaluation

james-powis:
the $::os fact hash has been around since 3.5 (and puppet 3 is EOL) this change removes the is_hash dependency provided by stdlib (which is deprecated). 

Supporting references:
https://github.com/puppetlabs/puppetlabs-stdlib#is_hash

Replacements for the is_ functions:
In param block: https://github.com/puppetlabs/puppetlabs-ntp/blob/master/manifests/init.pp#L17-L28
In code block: https://docs.puppet.com/puppet/4.8/lang_expressions.html#regex-or-data-type-match

Note: both of the replacements for is_* are only compatible with puppet 4.0.0 and later, I have updated the metadata.json to reflect this, and I would recommend releasing a new major release version to the forge.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -49,15 +49,16 @@ class gitlab::install {
         }
       }
       'redhat': {
-
         $gpgkey = $edition ? {
             'ee'    => 'https://packages.gitlab.com/gitlab/gitlab-ce/gpgkey https://packages.gitlab.com/gitlab/gitlab-ee/gpgkey/gitlab-gitlab-ee-3D645A26AB9FBD22.pub.gpg',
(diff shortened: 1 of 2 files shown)


tobru:
Thanks for your PR. We're not yet ready to deprecate Puppet 3 support, this will happen over the next weeks.

manifests/install.pp
```diff
@@ -50,11 +50,7 @@
         }
       }
       'redhat': {
-        if is_hash($::os) {
-          $releasever = $::os[release][major]
-        } else {
-          $releasever = $::operatingsystemmajrelease
-        }
+        $releasever = $::os[release][major]
```
bastelfreak: this should be $facts['os']['release']['major']

bastelfreak:
@james-powis thanks for the PR, can you please rebase?

james-powis:
@bastelfreak Done and done, we should be good to go... as $::operatingsystemmajrelease is legacy status, I reverted the changes which were introduced in 3fa7b17234c765e15c04b5f13aaf46b16710e27e which 80% covered this PR.

bastelfreak:
Thanks!

PR #110, opened 2017-01-21
Adds ability to specify git_data_dirs

logicminds:
* previously if you wanted to specify a data directory you would use
    git_data_dir.  However this setting was deprecated in 8.10 in favor
    of git_data_dirs which is a hash of multiple directories.

  * This adds git_data_dirs to the gitlab.rb template and populates
    the hash with the legacy git_data_dir if used, as well as also
    adding the legacy setting.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -16,6 +16,12 @@ class gitlab::config {
   $external_port = $::gitlab::external_port
   $git = $::gitlab::git
   $git_data_dir = $::gitlab::git_data_dir
+  # git_data_dirs is the new way to specify data_dirs, introduced in 8.10
+  if $git_data_dir {
+    $git_data_dirs = { 'default' => $::gitlab::git_data_dir }.merge($::gitlab::git_data_dirs) 
+  } else {
+    $git_data_dirs = $::gitlab::git_data_dirs
+  }
   $gitlab_git_http_server = $::gitlab::gitlab_git_http_server
   $gitlab_ci = $::gitlab::gitlab_ci
   $gitlab_pages = $::gitlab::gitlab_pages
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -300,6 +300,7 @@ class gitlab (
   $external_port = undef,
   $git = undef,
   $git_data_dir = undef,
+  $git_data_dirs = {},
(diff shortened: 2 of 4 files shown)


logicminds:
Please note this change may not work on gitlab before 8.10 as I am not sure how omnibus processes old settings or settings that don't exist yet.

The expected output will be the following when only using git_data_dir 

```
## THIS CONFIGURATION IS MANAGED BY PUPPET
## For setting up different data storing directory
## see https://gitlab.com/gitlab-org/omnibus-gitlab/blob/629def0a7a26e7c2326566f0758d4a27857b52a3/README.md#storing-git-data-in-an-alternative-directory
## If you want to use a single non-default directory to store git data use:
git_data_dir "/var/opt/data"

### For setting up different data storing directory
###! Docs: https://docs.gitlab.com/omnibus/settings/configuration.html#storing-git-data-in-an-alternative-directory
###! **If you want to use a single non-default directory to store git data use a
###!   path that doesn't contain symlinks.**
git_data_dirs({"default"=>"/var/opt/data"})

```  

If only using git_data_dirs than the output looks like:

```
## THIS CONFIGURATION IS MANAGED BY PUPPET
## For setting up different data storing directory
## see https://gitlab.com/gitlab-org/omnibus-
### For setting up different data storing directory
###! Docs: https://docs.gitlab.com/omnibus/settings/configuration.html#storing-git-data-in-an-alternative-directory
###! **If you want to use a single non-default directory to store git data use a
###!   path that doesn't contain symlinks.**
git_data_dirs({"default"=>"/var/opt/data"})

```

logicminds:
Docs about this config are here: https://docs.gitlab.com/omnibus/settings/configuration.html#storing-git-data-in-an-alternative-directory

tobru:
Thanks a lot for your contribution. I'd happy to merge it but unfortunately the test fails. Would you mind to fix either the test or your code before I do the merge?

tobru:
@logicminds I'd be happy to merge your PR, would you mind to rebase and check if there are no failing tests? I'll then merge your PR.

davidgubler:
I had to fix some things (a9bb637fcbd36f441ee3da013c7be93491c3da27), but I think it looks good now.

PR #111, opened 2017-01-25
Add settings for Prometheus

mansong1:


danjng:
Would be cool to roll this in! :)

Will give this commit a shot when I get a chance in my test environment.

PR #116, opened 2017-02-09
fixed method for sorting hashes in gitlab.rb to sort ldap hashes too

rwuest:
Fixed Method for sorting the hashes, so that the ldap hash can be sorted too. If the order of the hash elements in ldap changes at every puppet run, every time a reconfigure and a restart of the gitlab will be triggered.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## v.1.11.2 [2017-02-09]
+* Fixed Sorted Hash method in gitlab.rb template to sort ldap hash too
+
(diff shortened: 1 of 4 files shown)


LongLiveCHIEF:
Ironic.  I'm been having _exactly_ this problem, and didn't realize why until I just wandered by to see what PR's were up for this module.  I'm going to give this a try ASAP.

tobru:
I had to resolve some merge conflicts, hopefully I did it in the correct way...

rwuest:
cool thanks a lot!

PR #124, opened 2017-03-09
Add prometheus support

chriscowley:
Added a parameter, spec test and simple docs to enable the embedded Prometheus server that was added in Gitlab 8.16


tobru:
Already done in #111

PR #125, opened 2017-03-21
Do not decorate within hash, this leads to doubled quotes

andrekeller:
When we decorate within a hash, this leads to doubled quoting, i.e.

```
gitlab_ci['gitlab_server'] = {"app_id"=>"'xxxyyy'", "app_secret"=>"'zzzvvv'", "url"=>"'https://git.example.net'"}
```


tobru:
Afaics this is solved on the latest release

PR #142, opened 2017-06-26
Drop warning about RPM support

djjudas21:
I'm using this module on CentOS 6 and 7, and it works perfectly, so you can probably drop the warning. Your `metadata.json` is already correct.


davidgubler:
Thank you for the feedback and the pull request!

PR #147, opened 2017-07-11
Refactor cirunners

b4ldr:
Hi,

I recently took a look at implementing runners for gitlab ci so i looked to your great module.  however i was not a fan of the current implmentation.  
  * having to define root level hiera objects seemed messy
  * the lack of validation on the input to the runners config seemed risky.
As such here is a refactor patch.  Its a fairly big patch however most of the code is boiler plate stuff.  The deisgn of the module is that we now have 

  * class `gitlab::cirunner`
  * class `gitlab::cirunner::${executor}`
  * define `gitlab::runner::${executor}`

this allows us to set global defaults in class gitlab::cirunner.  executor specific defaults in gitlab::cirunner::${executor} and then override them in gitlab::runner::${executor} if needed.  A lot of the code relates to spec tests, however i think much of this could be dropped as long as we are happy with the function spec tests for `cmd_str`.

I have also written my updates with type specifiers and not bothered with `validate_*` functions.  This means that `PUPPET_VERSION=' ~> 3.8.0' FUTURE_PARSER=no` will no longer work.  I would also be happy to convert the rest of the module to use this syntax however: 
  1) this pull request is already quite big 
  2) I would prefer to rewrite things using the `puppetlabs-stdlib` type definitions e.g. `Stdlib::Unixpath`.  i'm not sure which version first supports these but puppet 4.3 did not.   Ideally it would be good to only support the (current-releases)[https://docs.puppet.com/release_notes/#current-releases] but thats obviously your decision, if you need to support older clients perhaps you could create a `future` branch which i can work from.

The other thing i did was to update your Gemfile.  i use https://github.com/voxpupuli/modulesync_config to help manage these files i would highly recommend you check it out.  anyway if you are not happy with changes made their let me know and ill remove them from the push.


tobru:
Hi @b4ldr. Thanks a lot for your PR. We'll review it soon(ish) and will let you know about it.

b4ldr:
Sure i appreciate its big so understand.  I would recomend starting with understanding how the `::gitlab::runner::shell` define works.  it uses `::gitlab::cirunner::cmd_str()` to create the cmd line arguments.  it uses similar code to the `pick_default` command  to work on the first valid value it finds then produces a comand line argument based on the value type.

if you follow how the shell command works then every thing else is just copy, paste, sed, repeat

b4ldr:
im not really happy with how i have named things.  i think one of the following options make sense

# everything under gitlab::runners:
```
cirunner.pp -> runners.pp
cirunner/shell.pp -> runners/shells.pp
```

# everything under cirunner
```
cirunner/shell.pp -> cirunner/shells.pp
runner/shell.pp -> cirunner/shell.pp
```

the first method would allow the current code to remain in place, which may be nice to allow transition

davidgubler:
Tagged "puppet4" due to the future parser requirement.

b4ldr:
close for #183

PR #148, opened 2017-07-11
Support runners setup via parameters on cirunner.pp

marcdeop:
Allow setting up of cirunners not onlly via hiera but also via parameters ( common_config and runners_config)

### DIFF
--- a/manifests/cirunner.pp
+++ b/manifests/cirunner.pp
@@ -19,6 +19,14 @@
 #   Default: gitlab_ci_runners
 #   Name of hiera hash with individual runners to be installed.
 #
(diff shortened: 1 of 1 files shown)


b4ldr:
I just raised a pull request against this functionality my self #147 however my changes are a bit more extensive.  that said as you are using the functionality i would be interested in your comments to my pull

LongLiveCHIEF:
After reviewing #183, I think that functionality supercedes the functionality here... however, after review, I think it makes more sense to extract this functionality into it's own module. The primary scope of _this_ module is to configure and install gitlab-omnibus.

PR #154, opened 2017-08-11
add support for chained global hooks

hboomsma:
> Introduced in GitLab Shell 4.1.0 and GitLab 8.15.
> 
> Hooks can be also placed in hooks/<hook_name>.d (global).


tobru:
Thanks

PR #155, opened 2017-08-18
Add backup job

b4ldr:
Small pull to add managing the gitlab backup cron job

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -62,6 +62,7 @@ class gitlab::config {
   $service_group = $::gitlab::service_group
   $service_manage = $::gitlab::service_manage
   $service_user = $::gitlab::service_user
+  $rake_exec = $::gitlab::rake_exec
   $shell = $::gitlab::shell
   $sidekiq = $::gitlab::sidekiq
   $sidekiq_cluster = $::gitlab::sidekiq_cluster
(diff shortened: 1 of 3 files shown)


LongLiveCHIEF:
This one has been on my list also, but I think it needs some things before it can be merged. Mainly, removing hardcoded values in favor of parameters.

b4ldr:
@LongLiveCHIEF have updated with params for minute and hour.  i also added a param for the gitlab-rake bin although this should not really change.  I dont think this needs any further paramertisation but feel free to suggest some

LongLiveCHIEF:
Those were the ones I was worried about in order to make it a good PR.  If you _really_ wanted bonus points, you could add SKIP's.  LGTM as is though.

LongLiveCHIEF:
Unfortunately the addition will cause a broken command syntax. Skips need
to be in a very specific unspaced, comma-sepatated format, and this prints
them out as an array.

The quick win here would be just to change the attribute type to a string
and allow entry of whatever is passed in, leaving it up to the implementer
to pass a valid value for skips.

The other option would be to have a parameter called other_args and just
append that to the backup cron.

b4ldr:
@LongLiveCHIEF you may have missed the join (https://github.com/vshn/puppet-gitlab/pull/155/files#diff-0ac866172efd29d0ddc2aa75ae6f182dR81) although it was joining them on white space before so i have fixed that to a comma

b4ldr:
Further to this the reason i prefer an array over a string is for data validation when we move to puppet4.  at that point we can do something like

# types/skip.pp
```puppet
type Gitlab::Skip = Enum['db', 'uploads', 'repositories', 'builds', 'artifacts', 'lfs', 'registry', 'pages']
```


# manifests/init.pp
```puppet
class gitlab {
 ...
 Array[Gitlab::Skip] $backup_cron_skips
 ...
}

LongLiveCHIEF:
Yeah, missed that. And speaking of skips... Puppet 5? Lol

b4ldr:
anything preventing this?

tobru:
Thanks

PR #156, opened 2017-09-07
Exact pattern match for the gitlab runner name in the toml config

fiddyspence:


tobru:
Could you have a look at the failing tests?

manifests/runner.pp
```diff
@@ -49,7 +49,7 @@
   # Execute gitlab ci multirunner register
   exec {"Register_runner_${title}":
     command => "/usr/bin/gitlab-ci-multi-runner register -n ${parameters_string}",
-    unless  => "/bin/grep ${runner_name} ${toml_file}",
```
bastelfreak: should we use https://github.com/puppetlabs/puppetlabs-stdlib#shell_join here?

LongLiveCHIEF:
I'm going to close. this would be superceded by #183, but I closed that PR because it made more sense to take all this functionality and move it into its own module.

Issue #157, opened 2017-09-26
Support new package repo


dandunckelman:
After the update to 10.x, the packages are now in a new repo & use a different package name (`gitlab-runner`).

https://docs.gitlab.com/runner/install/linux-repository.html#installing-the-runner

davidgubler:
Merged and released as v1.15.2. Thanks!

PR #160, opened 2017-09-27
Add 'package_name' param to cirunner class

dandunckelman:
Should fix https://github.com/vshn/puppet-gitlab/issues/157

### DIFF
--- a/manifests/cirunner.pp
+++ b/manifests/cirunner.pp
@@ -152,6 +152,7 @@ class gitlab::cirunner (
   $runners = keys($runners_hash)
   $default_config = hiera_hash($hiera_default_config_key, {})
   gitlab::runner { $runners:
+    binary         => $package_name,
     default_config => $default_config,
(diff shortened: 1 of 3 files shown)


dandunckelman:
I got bit by this:  https://gitlab.com/gitlab-org/gitlab-runner/issues/2757
- The `gitlab-runner` binary seems to be picky about the order of CLI options

So, I had to run the registration process manually.  If/when that gets resolved upstream, maybe changes will be needed for the params passed to the binary.

davidgubler:
Looks good to me (although I don't like the fact that the runner path must be manually set for Gitlab 10, but I see no easy way around that for now).

dandunckelman:
@davidgubler thanks.

You said:
> the runner path must be manually set for Gitlab 10

Are you talking about `/usr/bin/gitlab-runner` and the new `binary` param on the `runner` define?

PR #161, opened 2017-10-04
Fixed apt version dependencies

hp197:
Checked and tested with most current java module.
Works like a charm.


tobru:
Already merged in another PR

PR #169, opened 2017-12-01
apt dep version bump

minorOffense:
Apt changelog says it's fully backwards compatible. Figure just try a new version number in tests. And it appears the 3.x was skipped altogether.


tobru:
Thanks

PR #172, opened 2017-12-06
Fix repo URL for RHEL 7.  Enable gpgcheck and add in gitlab-ee key.

mhyzon:
The yum variable $releasever is '7' in centos, but '7Server' in RHEL.

Also adding the proper GPG based on the edition installed (CE is signed by a different key than EE).  I am not sure if there is a better or more canonical place for their GPG key(s)


ploeghvander:
look ok. please merge.

tobru:
Thanks

ploeghvander:
Thanks. When will it be released?

PR #175, opened 2017-12-18
fix typo

bc-bjoern:
The variable "glocal_hooks" does not exists, so it resulted in this error: 

`Error: "" is not a Hash.  It looks to be a String at /tmp/vagrant-puppet/modules-7ee4745f90860a506980a9923d6fea86/gitlab/manifests/init.pp:490 on node puppet.localdomain`


tobru:
Thanks!

PR #174, opened 2017-12-15
initial modulesync

tobru:
This is the first modulesync run:

```
msync update -m "initial modulesync" -b "modulesync" -f "puppet-gitlab"
```

PR #176, opened 2017-12-21
Fixes via rubocop -a

kallies:
@tobru here are the changes from `rubocop --auto-correct`(#174)


tobru:
Thanks. Merged into #174

PR #177, opened 2018-01-09
Db indexing for git authorized keys

### LINKED ISSUE #168: Feature: Allow database to index git authorized_keys
As described in the [speed up ssh operations](https://docs.gitlab.com/ee/administration/operations/speed_up_ssh.html), gitlab has a feature that will allow the gitlab_shell to index authorized_keys for git+ssh connections in the database instead of the `authorized_keys` file. This can drastically impact the speed of ssh connections when a large number of keys or users are present.

Implementation details:
- [x] Add `store_git_keys_in_db` parameter to `manifests/init.pp`
- [x] create `/opt/gitlab-shell/authorized_keys` keys file with `git:root 0650` with content shown in docs
- [ ] add `manage_sshd_git_user` parameter
- [ ] use `augeas` resource to configure sshd with appropriate command, or use sshd drop-in file for git user

LongLiveCHIEF:
This enhancement allows for fast lookups for ssh keys for communicating with gitlab instances using git+ssh protocol.

As described in the speed up ssh operations documentation, gitlab has a feature that will allow the gitlab_shell to index authorized_keys for git+ssh connections in the database instead of the authorized_keys file. This can drastically impact the speed of ssh connections when a large number of keys or users are present.

closes #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ the `$::os` fact used in `install.pp` doesn't work as expected.
 
 ### Beginning with Gitlab
 
-Just include the class and specify at least `external_url`. If `external_url` is not specified it will default to the FQDN fact of the system. 
+Just include the class and specify at least `external_url`. If `external_url` is not specified it will default to the FQDN fact of the system.
 
 ```puppet
 class { 'gitlab':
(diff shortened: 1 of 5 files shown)


LongLiveCHIEF:
so it looks like my builds are failing because there was a network timeout during one of the jobs. Only that one job failed, and all the others passed.  Looks like this is a common problem with ruby deps, according to Travis help docs: https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out. 

I'm going to close and re-open the PR just to see if it will try to rebuild.

PR #179, opened 2018-01-12
Update for rubocop compliance

mterzo:


mterzo:
I'm not that fluent with Travis-CI.  The branches are limited to only master and a version regex, so it looks like it won't run with my PR going to your branch.

```
Running RuboCop...
Inspecting 9 files
.........
9 files inspected, no offenses detected
```

PR #182, opened 2018-01-24
Added postgres upgrade tasks

Nekototori:
Was going through upgrading my older gitlab and had to upgrade the postgres database.  This task follows the steps and checks here:
https://docs.gitlab.com/omnibus/settings/database.html#upgrade-packaged-postgresql-server

I've broken it up into two.  One does the upgrade, and another does the cleanup.

### DIFF
--- a/tasks/post_upgrade.sh
+++ b/tasks/post_upgrade.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+echo 'Cleaning up leftover files from upgrade...'
+sudo rm -rf /var/opt/gitlab/postgresql/data.9.2.18
--- a/tasks/postgres_upgrade.sh
+++ b/tasks/postgres_upgrade.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+echo 'Checking pgsql version'
+CMD=$(sudo gitlab-psql --version)
+echo "version is ${CMD##* }"
+#9.2.18
+if [ "${CMD##* }" = "9.2.18" ]; then
+  echo 'Version is below required for Gitlab 10+, checking...'
+  DB_SIZE=$(du -shk /var/opt/gitlab/postgresql/data | awk '{print $1}')
+  FREE=$(df -hk /var/opt/gitlab/postgresql/data/ | tail -1 | awk '{print $4}')
+  echo "Database size is: $DB_SIZE kb and freespace is $FREE kb"
+  if [ $DB_SIZE -lt $FREE ]; then
+    echo 'Enough freespace available to proceed.'
+    sudo gitlab-ctl pg-upgrade
+    echo 'Upgrade complete.  Please verify everything is correct and then run the post_upgrade task.'
+  else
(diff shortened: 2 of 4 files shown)


tasks/postgres_upgrade.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+echo 'Checking pgsql version'
+CMD=$(sudo gitlab-psql --version)
+echo "version is ${CMD##* }"
+#9.2.18
+if [ "${CMD##* }" = "9.2.18" ]; then
```
eputnam: would it better to do a `-lt 10` here instead of hardcoding a 9.x?

LongLiveCHIEF: Not in this case. Refer to the link in the OP's PR.

tasks/postgres_upgrade.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+echo 'Checking pgsql version'
+CMD=$(sudo gitlab-psql --version)
```
LongLiveCHIEF: The `sudo` needs to come out here, and this task needs to be run with `--sudo --tty --run-as root`

tasks/postgres_upgrade.sh
```diff
@@ -0,0 +1,22 @@
...
+echo "version is ${CMD##* }"
+#9.2.18
+if [ "${CMD##* }" = "9.2.18" ]; then
+  echo 'Version is below required for Gitlab 10+, checking...'
+  DB_SIZE=$(du -shk /var/opt/gitlab/postgresql/data | awk '{print $1}')
+  FREE=$(df -hk /var/opt/gitlab/postgresql/data/ | tail -1 | awk '{print $4}')
+  echo "Database size is: $DB_SIZE kb and freespace is $FREE kb"
+  if [ $DB_SIZE -lt $FREE ]; then
+    echo 'Enough freespace available to proceed.'
+    sudo gitlab-ctl pg-upgrade
```
LongLiveCHIEF: remove `sudo`

tasks/post_upgrade.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+echo 'Cleaning up leftover files from upgrade...'
+sudo rm -rf /var/opt/gitlab/postgresql/data.9.2.18
```
LongLiveCHIEF: remove `sudo`

LongLiveCHIEF:
@Nekototori sorry it took so long to get to this. I had to familiarize myself with bolt and get used to using it with `sudo` before I could properly review this request.

I noted in three spots where your script uses `sudo` explicitly.  Tasks/scripts should not have `sudo` in them, and should instead be run with `--sudo --tty --run-as root` on the command line.

We also need to add some documentation to README, adding a `##Tasks` section, and then adding this as the first task.

Once you clean up the `sudo`'s, and add in some documentation for usage I'll merge. Make sure to mention the `--sudo --run-as root --tty` bit, since it's currently not documented fully in bolt docs.

Thanks for your contribution!

PR #184, opened 2018-01-25
Add requirement for puppet 4.6+

logicminds:
* this alone doesn't change any code but lets folks now we will
    be dropping support for puppet less than 4.7.  Why pupppet 4.7?
    Puppet 4.7 officially added support for datatypes which I think
    is the minimal version required.  Since stdlib provides many
    excellent datatypes we will need 4.7 support to load those external
    types.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -68,7 +68,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">=3.7.0 <5.0.0"
+      "version_requirement": ">= 4.7.0 <6.0.0"
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


logicminds:
Note: There are several PRs iin the queue to add datatype checking already.  Additionally, support for puppet 3.x is EOL.

metadata.json
```diff
@@ -68,7 +68,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">=3.7.0 <5.0.0"
+      "version_requirement": ">= 4.6.0 <5.0.0"
```
juniorsysadmin: Perhaps 4.7.1

logicminds:
Updated to 4.7.0.

PR #186, opened 2018-01-29
Add ha roles

LongLiveCHIEF:
This pull request adds the `roles` configuration item for controlling scaled instances of GitLab Omnibus.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -68,6 +68,7 @@ class gitlab::config {
   $gitlab_workhorse = $::gitlab::gitlab_workhorse
   $user = $::gitlab::user
   $web_server = $::gitlab::web_server
+  $roles = $::gitlab::roles
 
   # replicate $nginx to $mattermost_nginx if $mattermost_nginx_eq_nginx true
   if $mattermost_nginx_eq_nginx {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -251,6 +251,11 @@
 #   Default: false
 #   Replicate the registry Nginx config from the Gitlab Nginx config.
 #
+# [*roles*]
+#   Default: undef
+#   Array of roles when using a HA or Geo enabled GitLab configuration
+#   See: https://docs.gitlab.com/omnibus/roles/README.html for acceptable values
+#
 # [*secrets*]
 #   Default: undef
 #   Hash of values which will be placed into $secrets_file (by default /etc/gitlab/gitlab-secrets.json)
@@ -410,6 +415,7 @@ class gitlab (
   Optional[String]               $registry_external_url         = undef,
   Optional[Hash]                 $registry_nginx                = undef,
   Boolean                        $registry_nginx_eq_nginx       = false,
+  Optional[Array]                $roles                         = undef,
   Optional[Hash]                 $secrets                       = undef,
   Stdlib::Absolutepath           $secrets_file                  = $::gitlab::params::secrets_file,
   Optional[Hash]                 $sentinel                      = undef,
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -2,6 +2,7 @@
 sudo: false
 dist: trusty
 language: ruby
+
```
eputnam: :(

manifests/config.pp
```diff
@@ -75,6 +75,7 @@
   $backup_cron_enable = $::gitlab::backup_cron_enable
   $backup_cron_minute = $::gitlab::backup_cron_minute
   $backup_cron_hour = $::gitlab::backup_cron_hour
+  $roles = $::gitlab::roles
```
eputnam: is there a way to make this var name more descriptive? perhaps `$ha_roles` or something like that?

manifests/init.pp
```diff
@@ -247,6 +247,11 @@
 #   Default: false
 #   Replicate the registry Nginx config from the Gitlab Nginx config.
 #
+# [*roles*]
```
eputnam: docs! \o/

manifests/config.pp
LongLiveCHIEF: For the most part, this module uses variables that are consistent with the names of the config keys expected by the product. This allows users to reference the product documentation for how a config is expected to change the behavior of the product, and removes ambiguity/uncertainty by the implementor on whether or not _our_ name for a config is really _some other_ config for the [gitlab.rb.template](https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template), which is what this module configures.

manifests/init.pp
LongLiveCHIEF: see my comment above. The standard for this module is to document something that is _not_ a feature of the omnibus package configuration. Documenting every config would be redundant, and could cause unexpected problems when gitlab changes the behavior or possible sub-key values.

There fore, the documentation provided for this param is adequate and in keeping with the standards of this module.

LongLiveCHIEF:
@eputnam are you familiar with the product itself? It seems like some of the review items you raised are indicating that you're unaware that this module is basically puppetizing a chef configuration, and that we're merely automating the generation of that config file according to the docs maintained by the vendor.

We _can't_ be the source of documentation for these configs, because it creates too many issues when debugging configuration problems. When we automate some manual part of the omnibus package/service, _then_ we document it... otherwise we merely indicate an expected type for each param, and in the param docs we point to the usage of that config by the vendor.

It's confusing unless you're familiar with how GitLab Omnibus is administered.

.travis.yml
LongLiveCHIEF: Will fix.

LongLiveCHIEF:
@eputnam a good example of how we document non `gitlab.rb.template` features is actually present in one of my other open Merge Requests: [Db indexing for git authorized keys](https://github.com/voxpupuli/puppet-gitlab/pull/177/files).

This is a feature of GitLab that requires manual configuration by the installer, (whereas the majority are handled by the bundled `chef` process according to the settings realized in `gitlab.rb`). As such, I added a section to the `README.md` about enabling it in the module, in addition to  the param docs.

Make sense? Sound good? Maybe we should actually document this distinction in the README so that module consumers don't have to figure out implicitly how our documentation works?

LongLiveCHIEF:
Looks like I need to remove the params and template keys for:

```
redis_master_role
redis_slave_role
redis_sentinel_role
```
Those are no longer viable configs.

PR #205, opened 2018-03-14
Move backup to its own class

### LINKED ISSUE #204: Backup cron configuration will trigger gitlab restart

- Puppet: Every
- Ruby: Every
- Distribution: Every
- Module version: Current Master

Change any backup_cron_* paramerters configuration

## What are you seeing
Changes on backup cron will restart Gitlab

Changes on backup cron shouldn't restart Gitlab - only changes which are directly connected to the gitlab service should restart it

baurmatt:
This commit moves the backup cronjob to its own class. This ensures that
backup configuration changes do not trigger a Gitlab service restart.

Fixes #204.

### DIFF
--- a/manifests/backup.pp
+++ b/manifests/backup.pp
@@ -1,0 +1,24 @@
+# == Class gitlab::backup
+#
+# This class is called from gitlab for backup config.
+#
+class gitlab::backup {
+  $rake_exec = $::gitlab::rake_exec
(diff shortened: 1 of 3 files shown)


LongLiveCHIEF:
I think we still need to remove the backup/cron stuff from `config.pp` :smile:

baurmatt:
See, this happens if you work to early in the morning ;) Sorry :/

PR #207, opened 2018-03-28
Prepare for 2.0.0 

LongLiveCHIEF:
Decided to increment to next _major_ version because:

- officially dropped puppet 3 support by including param validation

Several other new minor features were also added, which would otherwise make this `1.17.0-rc0`.

See CHANGELOG.md in PR for details.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,6 +1,6 @@
 {
   "name": "puppet-gitlab",
-  "version": "1.16.2-rc0",
+  "version": "2.0.0-rc0",
   "author": "Vox Pupuli",
   "summary": "Installation and configuration of Gitlab Omnibus",
   "license": "BSD-3-Clause",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,32 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+## [v2.0.0](https://github.com/voxpupuli/puppet-gitlab/tree/v2.0.0) (2018-03-28)
+
+[Full Changelog](https://github.com/voxpupuli/puppet-gitlab/compare/v1.16.1...v2.0.0)
+
+**Merged pull requests:**
+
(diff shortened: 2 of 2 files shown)


LongLiveCHIEF:
@tobru would you mind reviewing this? I'm mostly concerned about the release process. The [Vox Pupuli "Releasing a new module version"](https://voxpupuli.org/docs/) documentation is a bit confusing, especially when it comes to whether or not the "release PR" should have `rc` identifiers or not in the `metadata.json` version tag.

If i follow the instructions, it wants us to merge this to master, and then travis will take over creating a tag. However, the docs don't indicate how travis would know to increment a major/minor version instead of a PATCH. 

I decided to use `2.0.0-rc0`, assuming this would be merged to master, and then I would run the `bundle exec travis_release` once it's merged.

You've done this release process once/twice now, so lmk if I'm doing it wrong.

Also, any feedback on the release would be great!

metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "puppet-gitlab",
-  "version": "1.16.2-rc0",
+  "version": "2.0.0-rc0",
```
juniorsysadmin: Set this to 2.0.0. When you do the `bundle release` after merging, the Rake task should do a commit to bump the version to the next version-rc0 in master.

LongLiveCHIEF:
That's what the docs said too, but it just doesn't make sense to me. *This*
commit is the release candidate for 2.0.0, and if the rake task bumps it to
the *next* rc0, then we'll never have a 2.0.0 and instead would skip right
to 2.0.1.

bastelfreak:
@LongLiveCHIEF the rake task wil create the git tag with the version from this PR, afterwards bump it to -rc0. The task verifies that the version it wants to release is a non rc one. This whole rc thing is just a protection for us to not accidently kick of a release.

bastelfreak:
Can you please rebase? We merged other changes to master during the past hours.

LongLiveCHIEF:
Alright, that makes sense.

> Can you please rebase? We merged other changes to master during the past hours.

No problem.

I also think I want to take care of https://github.com/voxpupuli/puppet-gitlab/issues/159 as part of the `2.0.0` release. I don't want to release a major version, and then release another major version a week or two later just for that one incompatibility.

PR #210, opened 2018-03-31
remove auto-execution of gitlab:setup

### LINKED ISSUE #35: 500 error on initial login?
Hi,

I just did a fresh install of Gitlab on RHEL 6.5 / Puppet 3.8.1 using the module, and I'm getting a 500 error when trying to access /users/sign_in (see below).  I found a lot of references to this type of error for upgrades / database migrations, but less for new installations.  I'm assuming that the "projects" table would be a default (since I don't have any data that I've added to gitlab yet)?

I'll keep digging, but I thought it was worth asking in case you've seen this error before after an install.

The production.log file says:

Started GET "/users/sign_in" for 127.0.0.1 at 2015-10-27 17:20:15 -0400
Processing by SessionsController#new as HTML
PG::UndefinedTable: ERROR:  relation "projects" does not exist
LINE 5:                WHERE a.attrelid = '"projects"'::regclass
                                          ^
:               SELECT a.attname, format_type(a.atttypid, a.atttypmod),
                     pg_get_expr(d.adbin, d.adrelid), a.attnotnull, a.atttypid, a.atttypmod
                FROM pg_attribute a LEFT JOIN pg_attrdef d
                  ON a.attrelid = d.adrelid AND a.attnum = d.adnum
               WHERE a.attrelid = '"projects"'::regclass
                 AND a.attnum > 0 AND NOT a.attisdropped
               ORDER BY a.attnum

Completed 500 Internal Server Error in 15ms (ActiveRecord: 9.7ms)

---

When I tried to troubleshoot by running gitlab-ci-rake commands, it complained about not being able to load files under /opt/gitlab/etc/gitlab-ci - that directory does not appear to exist.

[root@linvm2346 gitlab-rails]# gitlab-ci-rake db:migrate:status
/usr/bin/gitlab-ci-rake error: could not load /opt/gitlab/etc/gitlab-ci/gitlab-ci-rc
Either you are not allowed to read the file, or it does not exist yet.
You can generate it with:   sudo gitlab-ctl reconfigure

[root@linvm2346 gitlab-rails]# cd /opt/gitlab/etc
[root@linvm2346 etc]# ls -la
total 48
drwxrwxr-x 3 root root  4096 Oct 27 17:01 .
drwxrwxr-x 9 root root  4096 Oct 27 17:01 ..
drwx------ 3 git  root  4096 Oct 27 17:01 gitlab-rails
-rw-rw-r-- 1 root root 35191 Oct 27 08:43 gitlab.rb.template

I tried running gitlab-ctl reconfigure by hand to look for errors, etc., but there weren't any obvious issues (and it still didn't create the gitlab-ci directory).

### LINKED ISSUE #195: After upgrade to GitLab 10.4.3 each puppet run wants to remove

- Puppet: 3.8.7
- Ruby: ruby 2.0.0p648 (2015-12-16) [x86_64-linux]
- Distribution: CentOS Linux release 7.4.1708 (Core)
- Module version: 1.16.1

The same as before GitLab upgrade - see below.

## What are you seeing

After GitLab upgrade from 10.1.x to 10.4.3 each puppet run wants to remove postgres' internal_certificate and internal_key from gitlab-secrets.json and thus restart gitlab.

```
[root@XXXXXX ~]# puppet agent -t --noop
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Info: Caching catalog for XXXXXX.YYYYYY
Warning: Non-string values for the file mode property are deprecated. It must be a string, either a symbolic mode like 'o+w,a+r' or an octal representation like '0644' or '755'.
   (at /usr/share/ruby/vendor_ruby/puppet/type/file/mode.rb:69:in `block (2 levels) in <module:Puppet>')
Info: Applying configuration version '1518736992'
(...)
Notice: Class[Profiles::Common]: Would have triggered 'refresh' from 1 events
Notice: /Stage[main]/Gitlab::Config/File[/etc/gitlab/gitlab-secrets.json]/content:
--- /etc/gitlab/gitlab-secrets.json     2018-02-15 23:22:06.165707956 +0000
+++ /tmp/puppet-file20180215-134031-1focon      2018-02-15 23:24:28.530491292 +0000
@@ -27,9 +27,5 @@
     "gitlab_auth_endpoint": "https://XXXXXX.YYYYYY/oauth/authorize",
     "gitlab_token_endpoint": "https://XXXXXX.YYYYYY/oauth/token",
     "gitlab_user_api_endpoint": "https://XXXXXX.YYYYYY/api/v3/user"
-  },
-  "postgresql": {
-    "internal_certificate": "-----BEGIN CERTIFICATE-----(...redacted...)-----END CERTIFICATE-----\n",
-    "internal_key": "-----BEGIN RSA PRIVATE KEY-----(...redacted...)-----END RSA PRIVATE KEY-----\n"
   }
 }

Notice: /Stage[main]/Gitlab::Config/File[/etc/gitlab/gitlab-secrets.json]/content: current_value {md5}aaaaaaaaaa, should be {md5}bbbbbbbb (noop)
Info: /Stage[main]/Gitlab::Config/File[/etc/gitlab/gitlab-secrets.json]: Scheduling refresh of Exec[gitlab_reconfigure]
Notice: /Stage[main]/Gitlab::Config/Exec[gitlab_reconfigure]: Would have triggered 'refresh' from 1 events
Notice: Class[Gitlab::Config]: Would have triggered 'refresh' from 2 events
Info: Class[Gitlab::Config]: Scheduling refresh of Class[Gitlab::Service]
```

I disabled ssl to postgres with this in hiera and I have verified according to https://docs.gitlab.com/omnibus/settings/database.html that it works:
```
gitlab::postgresql:
  ssl: 'off'
```
...but it didn't help with the issue above.

I think that the problem is that GitLab generates those values of secrets file on each run, so even setting them in hiera doesn't seem to help.

Puppet run should not change gitlab-secrets.json and not restart gitlab.

## Any additional information you'd like to impart

We have been using old vshn/gitlab: 1.7.2 before switching to puppet-gitlab: 1.16.1 and it behaved the same way.

LongLiveCHIEF:
This code block was valid in older versions of gitlab, but in 10.x versions, there are several situations where you would be managing the gitlab service _and_ had configurations in the `postgresql` hash, _and_ have postgresql disabled, or merely not indicate and let the default psql config for the `enable` setting be utilized by omnibus chef.  (for example, gitlab HA configurations for database roles/nodes, or when providing your own psql database)

Additionally, the `git_data_dir` is no longer a valid param, and the postgres configuration doesn't use teh `setup.complete` file at this path any more. This causes gitlab to re-perform setup on each puppet run if you have postgres configurations but don't explictly indicate `enable => true`

Fixes #35 
Fixes #195

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -145,21 +145,6 @@ class gitlab::config {
       logoutput   => true,
       tries       => 5,
     }
-
-    if $postgresql =~ Hash {
-      unless $postgresql[enable] {
-        exec { 'gitlab_setup':
-          command     => "/bin/echo yes | ${rake_exec} gitlab:setup",
-          refreshonly => true,
-          timeout     => 1800,
-          require     => Exec['gitlab_reconfigure'],
-          unless      => "/bin/grep complete ${git_data_dir}/postgresql.setup",
-        }
-        -> file { "${git_data_dir}/postgresql.setup":
-          content => 'complete',
-        }
-      }
-    }
   }
 
   if $skip_auto_migrations != undef {


bastelfreak:
Thanks @LongLiveCHIEF ! Can you please create feature branches in a fork, and not in the org repo?

PR #219, opened 2018-04-16
init.pp: fix #217

### LINKED ISSUE #217: $web_server documentation differs from code
https://github.com/voxpupuli/puppet-gitlab/blob/f2ef16b98d2c892e931061595e280280e9b9c432/manifests/init.pp#L306-L308

https://github.com/voxpupuli/puppet-gitlab/blob/f2ef16b98d2c892e931061595e280280e9b9c432/manifests/init.pp#L431

`web_server` used to be of type hash, but now it is of type String. However, according to the documentation, it still is of type hash. So either the code or the documentation should be updated. I assume, that is is actually a bug in the code that appeared in commit 3fa7b17234c765e15c04b5f13aaf46b16710e27e

Same is also true for the `user` parameter.

NiklausHofer:
`$user` and `$web_server` are of type `Hash` rather than type `String`. 

Fixes #217


edransmpaz:
This fixes #220 as well.

PR #223, opened 2018-04-17
Fix wrong quotation in boolean

### LINKED ISSUE #222: Bad use of method 'inspect' causes bool values not to apply.
Based on gitlab documentation and feedback from support, I realized that the boolean values passed to the gitlab.rb file are not being parsed properly due to a misplaced quotes. Gitlab expects bool values to be passed as is, without any quotation.
Usage examples: 
postgres['enabled'] = 'false' -> Service still running visible via gitlab-ctl status
postgres['enabled'] = false -> Expected behaviour.

Main reason: https://ruby-doc.org/core-2.3.0/String.html#method-i-inspect

https://github.com/voxpupuli/puppet-gitlab/blob/1a1cc1e26644d7b1b8790115cf284632f2def525/templates/gitlab.rb.erb#L34

edransmpaz:
Case exception for Boolean values in decorator method inside gitlab.rb.erb template file.
Fixes #222

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -427,7 +427,7 @@ class gitlab (
   Boolean                        $store_git_keys_in_db          = false,
   Optional[Hash]                 $unicorn                       = undef,
   Optional[Hash]                 $gitlab_workhorse              = undef,
-  Optional[String]               $user                          = undef,
+  Optional[Hash]                 $user                          = undef,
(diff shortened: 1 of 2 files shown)


LongLiveCHIEF:
Merged. @edransmpaz thanks for the first contribution!

PR #243, opened 2018-06-08
Do not manage apt-transport-https.

jkroepke:
Hi,

since apt-transport-https is in Ubuntu 18.04/Debian 10 just a transitional package it isn't mandatory anymore, the apt-transport-https dependency should be conditional.

Since [puppetlabs-apt](https://github.com/puppetlabs/puppetlabs-apt/blob/master/manifests/source.pp#L32) 4.4.0 handle this correctly I prefer to consume that logic instead build a own.

See:

https://packages.ubuntu.com/bionic/apt-transport-https
https://packages.debian.org/buster/apt-transport-https

### DIFF
--- a/manifests/cirunner.pp
+++ b/manifests/cirunner.pp
@@ -72,7 +72,6 @@ class gitlab::cirunner (
     case $::osfamily {
       'Debian': {
         include apt
-        ensure_packages('apt-transport-https')
 
         apt::source { 'apt_gitlabci':
           comment  => 'GitlabCI Runner Repo',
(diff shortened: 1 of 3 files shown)


LongLiveCHIEF:
I have a refactor of repo/package management coming this weekend, and I'll make sure this is part of that.  It will be included in the next release, which should wind up being a major release (`v3.0.0)

LongLiveCHIEF:
lol, n/m, i thought this was an issue and not a PR. LGTM :thumbsup:

PR #254, opened 2018-07-09
fixed param inheritance in install manifest

### LINKED ISSUE #250: manage_package_repo is gone and replacement is broken
When you removed manage_package_repo - a doc was added https://github.com/voxpupuli/puppet-gitlab/commit/517469bd45e4a9ef61884550302b1080e53e2540#diff-04c6e90faac2675aa89e2176d2eec7d8R188 - stating how to use the new.
It does however NOT work.
I made it work by adding:
  Optional[String]               $package_name                  = undef,
  Boolean                        $manage_omnibus_repository     = true,
to gitlab class - and make package::install - default to using $gitlab::package_name and obmibus_.. class - to using $gitlab::manage_ominibus_repository

Otherwise - one cannot use gitlab - and have repos handled by other modules.

LongLiveCHIEF:
fixes #250

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ This approach provides the following advantages:
 In order to provide your own repository configurations, you are required to set `manage_upstream_edition => disabled`, and provide a hash
 of repository resource type configurations in the following format:
 
-```
+```yaml
 gitlab::repository_configuration:
   repository_resource_type: #ex... 'apt::source` or `apt::pin` or `yumrepo`
     repository_resource_title:
--- a/README.md
+++ b/README.md
@@ -139,10 +139,11 @@ In order to provide your own repository configurations, you are required to set 
 of repository resource type configurations in the following format:
 
 ```
-repository_resource_type: #ex... 'apt::source` or `apt::pin` or `yumrepo`
-  repository_resource_title:
-    repository_resource_attribute1: 'value'
-    repository_resource_attribute2: 'value'
+gitlab::repository_configuration:
+  repository_resource_type: #ex... 'apt::source` or `apt::pin` or `yumrepo`
+    repository_resource_title:
+      repository_resource_attribute1: 'value'
+      repository_resource_attribute2: 'value'
 ```
 
 Examples/defaults for `yumrepo` can be found at `data/RedHat.yaml`, and for `apt` at `data/Debian.yaml`.
@@ -179,17 +180,11 @@ As an expanded example of repository and package configuration, let's assume you
  - hosted inside your organizations firewall
  - installing gitlab-omnibus enterprise edition
 
-```
+```puppet
 class { 'gitlab':
   external_url => 'http://gitlab.mydomain.tld',
   manage_upstream_edition => 'disabled',
-}
-
-class { 'gitlab::install':
-  package_name => 'gitlab-ee'
-}
-
-class { 'gitlab::omnibus_package_repository':
+  package_name => 'gitlab-ee',
   repository_configuration => {
     'apt::source' => {
       'internal_mirror_of_gitlab_official_ce' => {
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -1,21 +1,10 @@
 # == Class gitlab::install
 #
 # This class is called from gitlab for install.
-#
-# [*package_name*]
-#   Default: 'gitlab-ce'
-#   The internal packaging system's name for the package
-#   This name will automatically be changed by the gitlab::edition parameter
-#   Can be overridden for the purposes of installing custom compiled version of gitlab-omnibus
-#
-# [*manage_package*]
-#   Default: true
-#   Should the GitLab package be managed?
-#
 class gitlab::install (
-  Optional[String] $package_name = undef,
+  $package_name = $gitlab::package_name,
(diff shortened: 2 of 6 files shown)


LongLiveCHIEF:
@KlavsKlavsen would you mind taking a look at this, since I know you've given this early preview a test already?

README.md
```diff
@@ -139,10 +139,11 @@ In order to provide your own repository configurations, you are required to set
 of repository resource type configurations in the following format:
 
 ```
-repository_resource_type: #ex... 'apt::source` or `apt::pin` or `yumrepo`
-  repository_resource_title:
-    repository_resource_attribute1: 'value'
-    repository_resource_attribute2: 'value'
+gitlab::repository_configuration:
```
bastelfreak: can you also add `yaml` behind the backticks, so we get proper syntax highlighting?

manifests/install.pp
```diff
@@ -1,21 +1,10 @@
...
-#   This name will automatically be changed by the gitlab::edition parameter
-#   Can be overridden for the purposes of installing custom compiled version of gitlab-omnibus
-#
-# [*manage_package*]
-#   Default: true
-#   Should the GitLab package be managed?
-#
 class gitlab::install (
-  Optional[String] $package_name = undef,
+  $package_name = $gitlab::package_name,
```
bastelfreak: can you please align the = chars?

manifests/install.pp
```diff
@@ -1,21 +1,10 @@
...
   $package_ensure = $gitlab::package_ensure,
-  Boolean $manage_package = true,
+  $manage_package = $gitlab::manage_package,
```
bastelfreak: I think everything that is not a private class, should have datatypes for parameters. I dont see a reason for removing them here. Can you explain why you did that? Even if it is private I would not remove the datatype, I would just not spend time implementing them. If this class is supposed to be private we should add a assert_private() call at the top

KlavsKlavsen:
@LongLiveCHIEF tested and it works great. All I had to do was to add manage_omnibus_repository => false - and it worked :)

README.md
LongLiveCHIEF: good call, will do!

manifests/install.pp
LongLiveCHIEF: Sure

manifests/install.pp
LongLiveCHIEF: I moved the type validation for this param into `manifests/init.pp`.  I've never seen the `assert_private()` functionality, but if keeps the class from being called from outside the module, then yes, this would be something I will add.

If I _do_ add `assert_private()`, then I'm assuming leaving data type validation in `manifests/init.pp` _only_ is fine, or should we still put the validation in both places?

LongLiveCHIEF:
@KlavsKlavsen I'm curious, if you're using an internal mirror or package repo for sourcing the gitlab omnibus package, why don't you leave `manage_omnibus_repository => true` and instead provide a configuration for `repository_configuration => {}` instead?

This was built for that exact use case, and provides some additional value in that scenario.

I'm just wondering if there's a implementation reason, or if you just weren't aware that's what this feature was designed for.  If it's the former, then say no more... I know how it goes with puppet control repo's and other modules you might have floating around that are tough to change. :smiley:

LongLiveCHIEF:
> Please have a look at the inline comments. Also I think it is a good idea to include this in the changelog. normally we only exclude release and modulesync PRs

@bastelfreak thanks for the great review. In regards to the changelog, this is a fix to the mechanics of PR #246, which _will_ be in the changelog. The summary of the intention behind _that_ PR in the changelog is what will be the important bit, as this will be a new major version release. 

I put the `skip-changelog` on this, because it's not fixing or enhancing anything that was ever officially included in a release version for this module. The param inheritance for `install` manifest isn't "broken" in the latest official release, and I didn't want to convey a message to module consumers that their currently deployed modules are broken.

manifests/install.pp
bastelfreak: I would still prefer it to keep the datatypes.

LongLiveCHIEF: I'm making development changes based on patterns i'm seeing in the rest of the VP modules, and I was just using the same standards I've seen elsewhere. I haven't seen any other examples where private classes have double verification. It's difficult for me to live up to your standards when they only apply to me.

https://github.com/voxpupuli/puppet-nginx/commit/206dc3d263b00ba9394aa6bf437f492431da51f3

KlavsKlavsen:
@LongLiveCHIEF  I could definetely use that feature.. but currently I have all "odd repos" collected under my own module - which handles them for both yum and apt repos.. so didn't want to bother with pulling one out, to use the feature (which is indeed a fine idea).

esalberg:
Is the expectation that everyone will need to specify a repository_configuration hash?  From reading the docs, it sounds like (and I agree) that a hash should not need to be set with manage_upstream_edition => 'ee' or 'ce' (i.e. not 'disabled')

if so, this parameter in init.pp needs a default - perhaps an empty hash?

Current:
Hash                           $repository_configuration,

New:
Hash                           $repository_configuration = {},

LongLiveCHIEF:
It _has_ a default. It's dynamic based on OS and provided via hiera in /data.

PR #255, opened 2018-07-11
Add gitlab-consul auth for pgbouncer database

### LINKED ISSUE #229: Support creation of .pgpass file
for gitlab HA installations, there is a `.pgpass` file that needs to be created on any `application_role` or `pgbouncer_role` nodes. Normally this is generated manually by running `gitlab-ctl write-pgpass....` command, which creates a file at the `userhome` for `gitlab-consul` user (default's to `/home/gitlab-consul`. 

This means we can use a template to create this file and automate this requirement just by having the file present.

- [ ] create variable for `pgpass_location`, defaults to `/home/gitlab-consul/.pgpass`
- [ ] create template for `.pgpass.erb`
- [ ] create variable for `pgbouncer_password` and inject it into `.pgpass.erb`

Optional/possibly required:
- make `host`, `database`, `user`, and `hostuser` template variables for the `.pgpass.erb` template 

The pattern for this file is such:

```
"${host}:*:${database}:${user}:${pgbouncer_password}
```

example outcome would be (with default values shown):

```
127.0.0.1:*:pgbouncer:pgbouncer:Password
```

LongLiveCHIEF:
This enhancement allows for the configuration of the authentication required for enabling built in database failover/service discovery. This PR adds configuration for the file the gitlab-omnibus package uses to authenticate the service discovery agent (consul) against the pgbouncer database.

closes #229

### DIFF
--- a/templates/.pgpass.erb
+++ b/templates/.pgpass.erb
@@ -1,0 +1,1 @@
+127.0.0.1:*:pgbouncer:pgbouncer:<%= @pgbouncer_password -%>
--- a/README.md
+++ b/README.md
@@ -386,6 +386,31 @@ You will need to configure the AuthorizedKeysCommand for the `git` user in sshd.
 Instructions for this are provided by GitLab at
 [Fast lookup of authorized SSH keys in the databasse](https://docs.gitlab.com/ee/administration/operations/fast_ssh_key_lookup.html)
 
+### Setting up GitLab HA
+
+#### pgbouncer Authentication
+
+For use in HA configurations, or when using postgres replication in a single-node setup, this module supports automated configuration
+of pgbouncer authentication. To set this up, set `pgpass_file_ensure => 'present'` and provide a valid value for `pgbouncer_password`.
+
+```puppet
+class {'gitlab':
+  pgpass_file_ensure => 'present',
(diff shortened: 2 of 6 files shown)


templates/.pgpass.erb
```diff
@@ -0,0 +1 @@
+127.0.0.1:*:pgbouncer:pgbouncer:<%= @pgbouncer_password -%>
```
bastelfreak: Can you please provide this an an epp template?
From our [guidelines](https://voxpupuli.org/docs/#reviewing-a-module-pr);
> Is a new template added? The preferred language is epp, not erb

Please provide the needed variable as [parameter](https://puppet.com/docs/puppet/5.5/lang_template_epp.html#accessing-variables).

bastelfreak:
Thanks for the enhancement @LongLiveCHIEF, can you please take a look at the inline comments I made?

templates/.pgpass.erb
LongLiveCHIEF: sure, will do.

LongLiveCHIEF:
@bastelfreak changed to `epp` and also added an additional test condition for template content that I forgot about. Ready to Pull.

PR #259, opened 2018-08-03
allow puppetlabs/apt 5.x

bastelfreak:


LongLiveCHIEF:
This dependency actually shouldn't be in the `metadata.json` at all.

PR #267, opened 2018-08-29
Allow configuration of the alertmanager via puppet.

### LINKED ISSUE #266: Alertmanager support in the omnibus config
Hi there,

We currently cannot configure the [alertmanager part of the omnibus config](
https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template#L1304).

Sample code:

```ruby
alertmanager => {
    flags => {
    'cluster.advertise-address' => "127.0.0.1:9093",
    }
}
```

Error seen in Puppet:

```
Class[Gitlab]: has no parameter named 'alertmanager'
```

gfokkema:
This PR adds support for configuring the Alertmanager gitlab service.
Documentation for that service is still quite sparse, best I've come across is:
https://prometheus.io/docs/alerting/alertmanager/

Right now our gitlab server is perpetually in error due to an unconfigured alertmanager,
and there's no way to disable it (this commit adds that).

Disabling alertmanager via hiera:
```
gitlab::alertmanager:
  enable: false
```

Fixes #266 

Couldn't find any unit tests for other params like this, so I haven't included them here either.
Please let me know :)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -60,6 +60,10 @@
 #   Default: /etc/gitlab/gitlab.rb
 #   Path of the Gitlab Omnibus config file.
 #
+# [*alertmanager*]
+#   Default: undef
+#   Hash of 'alertmanager' config parameters.
+#
 # [*ci_redis*]
 #   Default: undef
 #   Hash of 'ci_redis' config parameters.
@@ -364,6 +368,7 @@ class gitlab (
   String                         $service_group                   = 'root',
   # gitlab specific
   String                         $rake_exec                       = '/usr/bin/gitlab-rake',
+  Optional[Hash]                 $alertmanager                    = undef,
   Optional[Hash]                 $ci_redis                        = undef,
   Optional[Hash]                 $ci_unicorn                      = undef,
   Boolean                        $config_manage                   = true,
--- a/manifests/omnibus_config.pp
+++ b/manifests/omnibus_config.pp
@@ -8,6 +8,7 @@ class gitlab::omnibus_config (
 ){
 
   # get variables from the toplevel manifest for usage in the template
+  $alertmanager = $gitlab::alertmanager
   $ci_redis = $gitlab::ci_redis
   $ci_unicorn = $gitlab::ci_unicorn
   $external_url = $gitlab::external_url
--- a/spec/classes/init_spec.rb
+++ b/spec/classes/init_spec.rb
@@ -70,6 +70,20 @@ describe 'gitlab', type: :class do
               with_content(%r{^\s*nginx\['listen_port'\] = ('|)80('|)$})
           }
         end
+        describe 'alertmanager' do
+          let(:params) do
+            { alertmanager: {
+              'enable' => true,
+              'flags' => { 'cluster.advertise-address' => '127.0.0.1:9093' }
+            } }
+          end
+
+          it {
+            is_expected.to contain_file('/etc/gitlab/gitlab.rb'). \
+              with_content(%r{^\s*alertmanager\['enable'\] = true$}).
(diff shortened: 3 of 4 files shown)


manifests/init.pp
```diff
@@ -364,6 +368,7 @@
   String                         $service_group                   = 'root',
   # gitlab specific
   String                         $rake_exec                       = '/usr/bin/gitlab-rake',
+  Optional[Hash]                 $alertmanager                    = undef,
```
bastelfreak: Can this default to an empty hash please instead of undef?

From our docs at https://voxpupuli.org/docs/#reviewing-a-module-pr : 
> Are there new params with datatype Hash or Array? If possible, they should default to empty Hash/Array instead of undef. You can also enforce the datastructure like Array[String]

Can you please enforce the data structure?

bastelfreak:
Hi @gfokkema, thanks for the PR. Can you take a look at the used email address in the commit? It is not associated with your github accoun. Can you please also add a unit test for this?

gfokkema:
Dear @bastelfreak, in my opinion it's better to make changes according to the style of the file.
Should you want unit tests and empty hashes instead of undefs, I'd propose doing that in a separate PR for all Optional[Hash] values at once so the file stays consistent.

The same goes for unit tests regarding the Optional[Hash] values.

gfokkema:
Dear @bastelfreak I think it'd be more appropriate to label this as a bug as well.
This issue prevents a fresh gitlab install by puppet from starting correctly (gitlab-runsvdir).

manifests/init.pp
LongLiveCHIEF: `undef` is correct for this module. And empty hash would change the behavior of the installation. I believe you've been given this same feedback in several other PR's.

LongLiveCHIEF:
@gfokkema it's not really a bug. I classify a bug label to be equivalent with a piece of code in this module that doesn't work as intended.

As Gitlab continues to drop new releases each month, there are times when they add new configurations that outpace the configurations available in this module.

There's nothing _wrong_ with the code in this moudle (at least in regards to this PR), it just doesn't yet support the latest versions of GitLab Omnibus.  Therefore it's an enhancement, and not a bug.

When we add these, we need to make sure they don't cause errors with installations of older versions of GitLab as well, or that _could_ cause a bug.

LongLiveCHIEF:
> This issue prevents a fresh gitlab install by puppet from starting correctly via systemd (gitlab-runsvdir).
@gfokkema can you open an issue with the error reported by systemd? The error reported in #266 is a result of trying to utilize a parameter this module does not yet support, and has nothing to do with systemd.

LongLiveCHIEF:
@gfokkema thanks for the PR! This will be included in the next release ( `v3.0.0` ) of the `puppet-gitlab` module.

gfokkema:
Understood, thanks for the clarification.
> @gfokkema it's not really a bug. I classify a bug label to be equivalent with a piece of code in this module that doesn't work as intended.

manifests/init.pp
LongLiveCHIEF: @bastelfreak sorry.. there was undoubtedly a better way to say that. I didn't _mean_ to be such a jerk.

born4new:
@LongLiveCHIEF Any rough estimates on when the `v3.0.0` will be released?

LongLiveCHIEF:
First week of September I think. the last big bit was getting the puppet-gitlab_ci_runner functionality moved out into it's own module so that the first release of that module could coincide with the removal of those features from this module. That is mostly done, so I just need to make sure testing is tight on both modules and remove any deprecation code and it'll be good to go.

born4new:
@LongLiveCHIEF any new ETA for the V3 by any chance?

LongLiveCHIEF:
@born4new working on it this weekend. Got pulled onto some other stuff and delayed me from working on this. Sorry!

PR #271, opened 2018-09-14
 fix repositories missing the edition name

tequeter:
The current `master` branch would generate a `gitlab_official_` repository with an invalid URL. Basically, all edition names (ce/ee) were missing in the repository configuration.

The root of the issue was the attempt at accessing the Puppet catalog's `gitlab::edition` from Hiera. This can't work AFAIK unless users are instructed to only configure the classes through Hiera (undesirable IMO).

I fixed it by providing both CE/EE definitions in Hiera and adding the `ensure` attribute in Puppet to match the requested edition. By managing both repositories, this also ensures that the user cannot end up with a leftover. NB: the Puppet data manipulation step is uglier than it should be, by lack of a Hash -> Hash map function.

The first commit documents the bug, should you prefer a different solution.

### DIFF
--- a/manifests/omnibus_package_repository.pp
+++ b/manifests/omnibus_package_repository.pp
@@ -18,7 +18,31 @@ class gitlab::omnibus_package_repository (
       $_edition = $manage_upstream_edition
     }
 
-    $_repository_configuration = $repository_configuration
+    if $_edition == 'disabled' {
+      $_repository_configuration = $repository_configuration
+    } else {
+      # if we manage the repositories, adjust the ensure => present/absent
+      # attributes according to the desired edition.
+      $_repository_configuration = $repository_configuration.reduce({}) | $_memo, $_pair1 | {
+        # yumrepo =>  ...
+        [ $_rsc_type, $_repo_hash ] = $_pair1
+
+        $_mapped_repo_hash = $_repo_hash.reduce({}) | $_memo, $_pair2 | {
+          # gitlab_official_ce => ...
+          [ $_repo_name, $_repo_attrs, ] = $_pair2
+
+          if $_repo_name == "gitlab_official_${_edition}" {
+            $_ensure = 'present'
+          } else {
+            $_ensure = 'absent'
+          }
+
+          $_memo + { $_repo_name => $_repo_attrs + { ensure => $_ensure } }
+        }
+
+        $_memo + { $_rsc_type => $_mapped_repo_hash }
+      }
+    }
 
     # common attributes for all repository configuration resources
     # ensures correct ordering regardless of the number or configuration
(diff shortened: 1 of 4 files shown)


manifests/omnibus_package_repository.pp
```diff
@@ -18,35 +18,30 @@
...
-          }
-        }
-        'RedHat': {
-          $_filtered_repository_configuration = {
-            'yumrepo' =>  {
-              "gitlab_official_${_edition}" =>  {
-                baseurl => "https://packages.gitlab.com/gitlab/gitlab-${_edition}/el/${facts['os']['release']['major']}/\$basearch",
-              },
-            },
+    if $_edition == 'disabled' {
```
LongLiveCHIEF: This whole batch of code was planned to be removed prior to the 3.0 release. It's only in here so that those who use the master in their environments will have a deprecation period.

This isn't a review, just a comment FYI.  I probably should have left a comment to this effect in the code when this was refactored.

LongLiveCHIEF:
Can you open this as an issue? Adding a `data/common.yaml` with a `gitlab::edition:` key will solve this problem, and allow us to remove all the code you refactored entirely.

tequeter:
I created bug #273 as requested.

I don't understand your planned fix, though. Would you mind detailing it?

AFAIK as long as you `%{lookup()}` a value from Hiera, you can't access class parameters set in the Puppet code (typically in the profile). I wrote a [failing testcase](https://github.com/tequeter/puppet-gitlab/blob/08b9b784e536a01646d4f278a7bded25a9b89b74/spec/classes/init_spec.rb#L44) illustrating this.

Are you implying that the users of the upcoming 3.0 version should stop configuring the `gitlab` class from Puppet and set some/all parameters through Hiera exclusively? I would find this inconvenient.

LongLiveCHIEF:
> AFAIK as long as you %{lookup()} a value from Hiera, you can't access class parameters set in the Puppet code (typically in the profile). I wrote a failing testcase illustrating this.

I don't know what I was thinking when I wrote that lookup in there, as that's not even how hiera works. 

> Are you implying that the users of the upcoming 3.0 version should stop configuring the gitlab class from Puppet and set some/all parameters through Hiera exclusively? I would find this inconvenient.

That was never my intent, and I don't even think this could be done.

I need a bit to think through what I was planning on having this file look like for the 3.0 release compared to this PR. Originally, I was going to remove the entire block coercing edition into the values we want, and it was going to look like this:

```
  if $manage_omnibus_repository {
    # common attributes for all repository configuration resources
    # ensures correct ordering regardless of the number or configuration
    # of repository related resources
    $resource_defaults = {
      tag    => 'gitlab_omnibus_repository_resource',
      before => Class['gitlab::install'],
    }

    # create all the repository resources
    $_repository_configuration.each() | String $resource_type, Hash $resources | {
      create_resources($resource_type, $resources, $resource_defaults)
    }
  }
```

That was going to be the extent of this manifest when I had planned it for the v3 release. Your PR though  has a few things I want to look at in conjunction with the `manifests/install.pp` and the `manage_upstream_edition` class parameter that was meant to replace the `edition` param.

PR #274, opened 2018-09-16
allow puppetlabs/apt 6.x;  require at least stdlib 4.13.1

bastelfreak:


LongLiveCHIEF:
`apt` is a soft dependency. It shouldn't be listed in `metadata.json` at all.

PR #272, opened 2018-09-14
Update gpgkey for yum repos

tequeter:
This updates the `gpgkey` line of the yum repositories as per Gitlab's install docs. At the moment, the `master` branch fails to install because of this. 

The first three commits are PR #271, 227abb5 is the only relevant change here.

I'll rebase this PR if needed when #271 is updated.

PR #281, opened 2018-10-17
fix the repositories as per #271 and #272

tequeter:
Somehow, after the #271 and #272 merges we ended up without the GPG key change for gitlab-ce and with `%{lookup()}` calls left in gitlab-ee. This commit fixes the data and adds tests for it.


bastelfreak:
Thanks @tequeter !

PR #283, opened 2018-10-18
do not require service if it's not being managed

zeromind:
Set appropriate value for the _require_ parameter of the _gitlab_reconfigure_ exec resource depending on whether the service is being managed.

Dependency error when _gitlab::service::service_manage_ is set to _false_

>  Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Could not find resource 'Service[gitlab-runsvdir]' in parameter 'require' (file: /etc/puppetlabs/code/environments/production/external_modules/gitlab/manifests/service.pp, line: 58) on node gitlab.local

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -18,6 +18,7 @@ class gitlab::service (
     $start = "${service_exec} start"
     $stop = "${service_exec} stop"
     $status = "${service_exec} status"
+    $reconfigure_attributes_require = [Service[$service_name], Class['gitlab::install']]
 
     service { $service_name:
       ensure     => $service_ensure,
@@ -29,6 +30,8 @@ class gitlab::service (
       hasstatus  => true,
       hasrestart => true,
     }
+  } else {
+    $reconfigure_attributes_require = Class['gitlab::install']
   }
 
   $reconfigure_attributes = {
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @zeromind. thanks for the PR! Can you add a unit test for this?

LongLiveCHIEF:
@zeromind can you open this as an issue as well please?

LongLiveCHIEF:
The more I think on this, the more I realize that the reconfigure doesn't require the service in order to be executed, and didn't need to require that resource to begin with.

I think this could be done much simpler by removing the `require` metaparameter from the `$reconfigure_attributes` hash, and adding a `require` with the correct resources in the if/else at the bottom of the `service.pp`, so that all the exec logic and resource ordering is handled in one place.

zeromind:
@LongLiveCHIEF If I understand correctly, the require parameter should not contain the service in any case, but always `Class['gitlab::install']`, which could be added to the initial `$reconfigure_attributes` hash.
Or should it depend on the service, but only if it is being managed?

LongLiveCHIEF:
The actual `gitlab-ctl reconfigure` will manage the gitlab service as needed during the run, and having puppet do anything to the service on a refresh signal will actually cause problems. Even _if_ you want to manage the service itself, the _reconfigure_ shouldn't be touching it.

alexjfisher:
The whole class might benefit from refactoring for readability?

LongLiveCHIEF:
@alexjfisher I don't think readability is the problem, so much as understanding how omnibus works with chef, and how that makes things so different than a normal puppet service class.

There are almost _no_ situations in which this module should actually touch the gitlab service itself directly, since that service is designed to be managed internally whenever a `gitlab-ctl` command is used.  Any direct manipulation of the service will break the operational status and even the ability to start or restart correctly afterwards.

PR #291, opened 2018-12-04
modulesync 2.4.1

bastelfreak:
modulesync 2.3.1

PR #290, opened 2018-12-04
Allow to set numeric values as string

smortex:
#### Pull Request (PR) description
When the user sets a value to the string representation of a number (e.g. `"25"`), the current code turn this into the number itself (`25` in this case).

#### This Pull Request (PR) fixes the following issues
Fixes #289

PR #292, opened 2018-12-13
Fix Continuous Integration

smortex:

Travis CI is currently broken.  Fix some issues found while inspecting the situation and attempting to unbreak CI.

No actual issues fixed, but it should help making #290 and #291 pass.

### DIFF
--- a/spec/acceptance/gitlab_spec.rb
+++ b/spec/acceptance/gitlab_spec.rb
@@ -20,6 +20,7 @@ describe 'gitlab class' do
     end
 
     describe command('curl -s -S http://127.0.0.1:80/users/sign_in') do
+      its(:exit_status) { is_expected.to eq 0 }
       its(:stdout) { is_expected.to match %r{.*reset_password_token=.*redirected.*} }
     end
   end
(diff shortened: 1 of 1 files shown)


Dan33l:
@smortex ia ora na.

Thank you for this PR. It looks that only puppet6-nightly fail. I don't understand why. I'll make some checks locally as soon as possible.

Dan33l:
@smortex this PR is interesting with `sleep` at right place and URL correctly called. Locally i ran it several times with centos7 and puppet6-nightly without any fail.

In the last push, job in Travis failed because of true fail, probably linked to network issue or something like this.

I asked to people with admin power in travis to rerun manually this uniq failing job. It should pass.
If so, i agree to merge the PR as is.

smortex:
@Dan33l sure. I can also force commit or squash commits so that Travis re-schedule a build, but it takes about 2 hours to run a build…

I am not sure the sleep really helps here: as far as I can see, the build failures where the same when the sleep was running too early.  Maybe it can be removed?

Dan33l:
Where it was previously in the code, the `sleep` was executed before the pp code is applied. Now, where it is located, it looks to be applied after the pp code is applied. Gitlab is a bundle of several softwares, so it should be interesting to wait all process are started before to test the result.

smortex:
For sure it does not harm, we can leave it as it is now.  :+1:

PR #293, opened 2018-12-15
remove service dependency cycle

### LINKED ISSUE #284: Dependency error when gitlab::service::service_manage is set to false
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Could not find resource 'Service[gitlab-runsvdir]' in parameter 'require' (file: /etc/puppetlabs/code/environments/production/external_modules/gitlab/manifests/service.pp, line: 58) on node gitlab.local
```

LongLiveCHIEF:
fixes #284

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -361,7 +361,7 @@ class gitlab (
   # system service configuration
   Boolean                        $service_enable                  = true,
   Enum['stopped', 'false', 'running', 'true'] $service_ensure     = 'running', # lint:ignore:quoted_booleans
-  Boolean                        $service_manage                  = true,
+  Boolean                        $service_manage                  = false,
   String                         $service_name                    = 'gitlab-runsvdir',
   String                         $service_exec                    = '/usr/bin/gitlab-ctl',
   String                         $service_user                    = 'root',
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -361,7 +361,7 @@
   # system service configuration
   Boolean                        $service_enable                  = true,
   Enum['stopped', 'false', 'running', 'true'] $service_ensure     = 'running', # lint:ignore:quoted_booleans
-  Boolean                        $service_manage                  = true,
+  Boolean                        $service_manage                  = false,
```
bastelfreak: Is the change of the default value required? That would be a backwards incompatible change and require a major release.

LongLiveCHIEF: Yeah, it is. See comments in #283.

This may be one of those weird cases where it may not require a major version change though. Even though it appears puppet was managing this service, it never really has been since the release of 2.x.  It's been managed by the `gitlab-cl reconfigure` process that gets kicked by install or configuration changes. There would be any impact to implementors with this change.

LongLiveCHIEF:
@bastelfreak let me know if my response makes sense. Coincidentally, this is also why no new tests were needed.

LongLiveCHIEF:
actually, i do need to make some unit test changes. (service expectations).  Still on the fence about the major version change, and would like input from contributors on that bit.

bastelfreak:
Thanks @LongLiveCHIEF !

PR #294, opened 2018-12-17
updating module descr and support info

LongLiveCHIEF:
The gitlab support information in the README hasn't changed since the very first release. I've added support guidelines for bugs and PR's as they relate to the versions of this puppet module, as well as the gitlab-omnibus product itself.

I also updated the old statement about gitlab 7.10, indicating that if you are using a gitlab package that is _not_ supported by gitlab, then older releases of this module may work better for you.

I tried to make it clear the circumstances for which PR's would be accepted as they relate to breaking or implementing deprecated features of the non-EoL versions of GitLab.

Hopefully this will help clarify the proper usage of this module and what support can be expected based on the version of gitlab-omnibus selected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,24 @@ Chef to configure all the services.*
 
 Supported are Debian based (Ubuntu, Debian) and RedHat based (CentOS, RHEL) operating systems.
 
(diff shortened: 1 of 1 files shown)


bastelfreak:
Thanks for the updates to the README.md!

PR #306, opened 2019-04-09
Ensure yumrepo contains description

clement-parisot:
Signed-off-by: Clément Parisot <clement.parisot@uni.lu>

When using this module under Redhat/Centos, the yum repo configuration file doesn't contain any "name" field. It generates some warnings for example when using yum-cron to update the packages:

```
/etc/cron.hourly/0yum-hourly.cron:

Repository 'gitlab_official_ee' is missing name in configuration, using id
```
On `yum-repo`puppet resource, *name* is addressed as [*descr* parameter](https://puppet.com/docs/puppet/5.4/types/yumrepo.html#yumrepo-attribute-descr). I chose to put the default value that we find when installing Gitlab from the official CE and EE repositories: `gitlab_gitlab-ce` and `gitlab_gitlab-ee`.

I haven't opened any issue to report this problem.

### DIFF
--- a/data/family/RedHat.yaml
+++ b/data/family/RedHat.yaml
@@ -2,6 +2,7 @@ gitlab::repository_configuration:
   yumrepo:
     "gitlab_official_ce":
       ensure: 'present'
+      descr: 'gitlab_gitlab-ce'
       assumeyes: true
       enabled: 1
       baseurl: "https://packages.gitlab.com/gitlab/gitlab-ce/el/%{facts.os.release.major}/$basearch"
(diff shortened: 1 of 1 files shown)


LongLiveCHIEF:
Closing for a few reasons. The base hiera data for this intentionally is as thin as possible so that end users can ensure the resulting repo definitions make sense for them. This means the repo definition we use has only the minimum required fields to allow yum to install the package.

The way we have implemented this, the `descr` can easily be provided as a parameter by the implementor, and is already supported. 

Finally, GitLab is in the process of merging ce and ee sources into a single package, and we would have to pull this change out shortly after putting it in, which could impact any downstream acceptance tests implementors write for their usage of the module.

Issue #189, opened 2018-02-03
Reformat registration options in --option=value output


mjburling:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.10.9
- Ruby: ruby 2.0.0p648 (2015-12-16) [x86_64-linux]
- Distribution: CentOS 7.4
- Module version: Master branch, v1.16.0 [2017-12-13]

## How to reproduce (e.g Puppet code you use)

```
# excerpt from from hieradata
gitlab_ci_runners_defaults:
  docker-allowed-images:
    - docker.internalnet.net/library/golang:*
    - docker.internalnet.net/library/python:*
    - docker.internalnet.net/rbh/*
  docker-extra-hosts:
    - hub.docker.io:127.0.0.1
  docker-image: docker.internalnet.net/rbh/java-8
  docker-volumes:
    - /cache
    - /home/gitlab-runner/.aws:/root/.aws:ro
    - /home/gitlab-runner/.gnupg:/root/.gnupg:ro
    - /home/gitlab-runner/.gradle/caches:/root/.gradle/caches:rw
    - /home/gitlab-runner/.gradle/gradle.properties:/root/.gradle/gradle.properties:ro
    - /home/gitlab-runner/.grails:/root/.grails:rw
    - /home/gitlab-runner/.ivy2:/root/.ivy2:rw
    - /home/gitlab-runner/.m2/repository:/root/.m2/repository:rw
    - /home/gitlab-runner/.m2/settings.xml:/root/.m2/settings.xml:ro
    - /home/gitlab-runner/.npm:/root/.npm:rw
    - /home/gitlab-runner/.npmrc:/root/.npmrc:ro
    - /home/gitlab-runner/.pulp/admin.conf:/root/.pulp/admin.conf:ro
    - /home/gitlab-runner/.pulp/uploads:/home/gitlab-runner/.pulp/uploads:rw
    - /home/gitlab-runner/.rbh:/root/.rbh:ro
    - /home/gitlab-runner/.sbt:/root/.sbt:rw
    - /home/gitlab-runner/.yarn:/root/.yarn:rw
    - /root/.ssh:/root/.ssh:ro
  executor: docker
  locked: false
[...]

PR #310, opened 2019-04-26
Separate data from code and move defaults to common.yaml

dhoppe:


LongLiveCHIEF:
Only parameters that _have_ to be configured by hiera (subject to hierarchical constraints) should be in hiera for a component module. This reduces complexity across versions of puppet and makes it easier for users to implement the module and rule out hiera as a source of bugs in their deployments.

alexjfisher:
Reopening as I think this needs further discussion and hopefully we can get some sort of consensus.

Dan33l:
IMO after consensus is found, it looks useful to add it in the guide line https://voxpupuli.org/docs/#reviewing-a-module-pr.

alexjfisher:
Maybe we *should* be following https://github.com/puppetlabs/best-practices/pull/24
This is quite a recent recommendation.  [`puppetlabs/ntp`](https://github.com/puppetlabs/puppetlabs-ntp/blob/b4dee625b2d03589ec6213ed5c087d712fddb1e8/manifests/init.pp) hasn't actually been updated.

Dan33l:
With PR  [voxpupuli.github.io #189](https://github.com/voxpupuli/voxpupuli.github.io/pull/189), i suppose we can cose this one.
@dhoppe  ?

PR #315, opened 2019-05-16
fix mistake with name variable git_data_dirs

tipmg:


LongLiveCHIEF:
This is actually correct as written. This isn't a configuration for gitlab, this is configuration for a git hook. the `git_dat_dir` is in reference to the path the hook should be placed on, which isn't the same as the `gitlab::git_data_dirs` value (which is actually a `Hash` of named paths).

PR #320, opened 2019-08-02
Support puppetlabs/stdlib 6.x

pillarsdotnet:


metadata.json
```diff
@@ -14,7 +14,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 6.0.0"
+      "version_requirement": ">= 4.13.1 < 7.0.0"
```
Dan33l: We use title of issues and PR to generate automatically `CHANGELOG.md`.
Can you change the title in a way to understand the change made ?

pillarsdotnet: Done.

Dan33l: Is it accurate enough for a reader of the `CHANGELOG.md` file?
6.0+ can be understood as 6.0 and higher. So 7.0?
Something like `puppetlabs/stdlib 6.x` was already used in others modules. There is no ambiguity.

thank you for PR @pillarsdotnet

PR #331, opened 2020-03-28
Use new gitlab gpg keys for package management

jkroepke:
As document in https://docs.gitlab.com/omnibus/update/package_signatures gitlab gpg keys expire soon and needs to be replace with new one.

The old key is valid until 2020-04-15 while is new keys starts at 2020-04-06.

A release of this module in this time slot should useful to avoid some local breaking changes.

This PR is created at 2020-03-28. I except test failures until 2020-04-06.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ class { 'gitlab':
         'comment' => 'Internal mirror of upstream gitlab package repository',
         'location' => 'https://my.internal.url/repository/packages.gitlab.com/gitlab/gitlab-ce/debian',
         'key' => {
-          'id' => '1A4C919DB987D435939638B914219A96E15E78F4',
+          'id' => 'F6403F6544A38863DAA0B6E03F01618A51312F3F',
           'source' => 'https://my.internal.url/repository/package.gitlab.com/gpg.key'
         }
       },
--- a/data/family/Debian.yaml
+++ b/data/family/Debian.yaml
@@ -4,11 +4,11 @@ gitlab::repository_configuration:
       comment: 'Official repository for GitLab Omnibus'
       location: "https://packages.gitlab.com/gitlab/gitlab-ce/debian"
       key:
-        id: '1A4C919DB987D435939638B914219A96E15E78F4'
+        id: 'F6403F6544A38863DAA0B6E03F01618A51312F3F'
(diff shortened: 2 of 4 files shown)


dhoppe:
@jkroepke You also need to cover the RSpec checks for RHEL, since the URL has changed.

jkroepke:
Unit tests are green now. ITs should work in 3 days.

baurmatt:
I don't like the fact that we have to wait until the key is active (and the old one is inactive) to merge/release this. What do you think about adding both keys and removing the old one in a couple of days/weeks/months?

dhoppe:
@baurmatt Just take a look at #333. @dnsmichi explained how to change the configuration manually. It should not be a problem to merge this pull request on time and create a new release.

dhoppe:
@baurmatt You do not need to reopen this pull request, to trigger an Travis CI build. The result will not change, because GitLab still offers the old GPG key. I already contacted the support via Twitter.

baurmatt:
> @baurmatt You do not need to reopen this pull request, to trigger an Travis CI build. The result will not change, because GitLab still offers the old GPG key. I already contacted the support via Twitter.

Interesting, that explains the failing test ;)

jkroepke:
Since gitlab is a global team, they should always announce dates with timezones ...

dnsmichi:
It is hard to track down when things are exactly done, especially when there's a certain chance that assignees change and so will timezones then. I can understand your concerns too, still, not so easy ;)

I'd recommend to follow the infrastructure group for these kind of changes. Luckily everything is developed and discussed in the open :) I just checked in there too, no changes to the issue yet.

https://gitlab.com/gitlab-com/gl-infra/infrastructure/-/issues/9576

dhoppe:
@JanKoppe You can trigger a rebuild after logging into the Travis CI web interface. I already did that, but there is still an issue regarding CentOS.

jkroepke:
Looking at 

https://packages.gitlab.com/install/repositories/gitlab/gitlab-ce/config_file.repo?os=centos&dist=7&source=script

the yum config doesn't need to be change since the gpg key redirects to the new location:

```
curl -sI https://packages.gitlab.com/gitlab/gitlab-ce/gpgkey | grep Location
Location: https://packages.gitlab.com/gpg.key
```

PR #338, opened 2020-04-08
Add support for running Gitlab with puma

### LINKED ISSUE #337: Support puma
Gitlab support running it's Rails application with Puma instead of Unicorn. This stable in the Omnibus package since 12.9.

https://about.gitlab.com/releases/2020/03/22/gitlab-12-9-released/#reduced-memory-consumption-of-gitlab-with-puma

We should support it.

baurmatt:
Fixes #337.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -302,6 +302,10 @@
 #   Default: undef
 #   Hash of 'unicorn' config parameters.
 #
+# [*puma*]
+#   Default: undef
+#   Hash of 'puma' config parameters.
+#
 # [*user*]
 #   Default: undef
 #   Hash of 'user' config parameters.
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -451,6 +455,7 @@
   Optional[Stdlib::Absolutepath] $source_config_file              = undef,
   Boolean                        $store_git_keys_in_db            = false,
   Optional[Hash]                 $unicorn                         = undef,
+  Optional[Hash]                 $puma                            = undef,
```
bastelfreak, line 458: is there any chance to enforce the datatype for the hash key and value?

baurmatt: Not really, because we would have to keep up with the changes done in the Omnibus package/Chef cookbook.

manifests/init.pp
```diff
@@ -302,6 +302,10 @@
 #   Default: undef
 #   Hash of 'unicorn' config parameters.
 #
+# [*puma*]
```
bastelfreak, line 305: we really need to convert this to puppet-strings :sadface:

baurmatt: https://github.com/voxpupuli/puppet-gitlab/issues/339 ;)

trenta:
Can a new version be pushed to the forge so we can use this?

bastelfreak:
@trenta thanks for the hint, I prepared a new release in https://github.com/voxpupuli/puppet-gitlab/pull/348. just waiting for approval.

PR #342, opened 2020-04-27
Add support for CentOS 8

dhoppe:


dhoppe:
Looks like there is an issue during the import / verification of the new GPG key.

lnemsick-simp:
@dhoppe Any idea when this will be resolved? Would like to see el8 support added!

trevor-vaughan:
I would like this as well!

op-ct:
:+1: 

CentOS 8 support would be great; it's been out for a year already.  Are the GPG failures still an issue?

dhoppe:
I just restarted the jobs and it looks like the problem has solved itself. 
* https://travis-ci.org/github/voxpupuli/puppet-gitlab/jobs/680048254
* https://travis-ci.org/github/voxpupuli/puppet-gitlab/jobs/680048255

Unfortunately we have to wait for Travis CI now, hoping that all jobs are green.

.travis.yml
```diff
@@ -26,59 +26,67 @@ jobs:
...
       services: docker
     - rvm: 2.5.3
       bundler_args: --without development release
-      env: BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_setfile=debian10-64 CHECK=beaker
+      env: BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_setfile=centos8-64 CHECK=beaker
       services: docker
     - rvm: 2.5.3
       bundler_args: --without development release
-      env: BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_setfile=ubuntu1604-64 CHECK=beaker
+      env: BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_setfile=debian8-64 CHECK=beaker
```
kenyon, line 53: Might as well remove this since Debian 8 is EOL.

dhoppe:
I agree, but this should be handled by a different pull request.

.travis.yml
dhoppe: I agree, but this should be handled by a different pull request.

bastelfreak (approved):

PR #369, opened 2020-11-26
Fix cut and paste error in documentation

sgnl05:


igalic (commented):
this file is generated, so this needs to be fixed at the source.

PR #375, opened 2021-01-12
Add support for the 'package' omnibus configuration hash.

rayward:

Allow the `package` hash of options - https://gitlab.com/gitlab-org/omnibus-gitlab/blob/0eed527687a2d78ef2efb8f86435acc02a3b6a97/files/gitlab-config-template/gitlab.rb.template#L2150 - to be configured so that options such as `systemd_after` and `systemd_wanted_by` can be defined.

N/A


rayward:
The CentOS builds are failing as mentioned here - https://github.com/voxpupuli/puppet-gitlab/pull/372#issuecomment-747377805

manifests/init.pp
```diff
@@ -130,6 +131,7 @@
   Optional[Hash]                 $high_availability               = undef,
   Optional[Hash]                 $logging                         = undef,
   Optional[Hash]                 $letsencrypt                     = undef,
+  Optional[Hash]                 $package                         = undef,
```
bastelfreak, line 134: can you enforce the datatype a bit? What about this:
```suggestion
  Hash[String[1],String[1]]      $package                         = {},
```

if possible, the default for arrays and hashes should always be an empty hash/array

rayward: I've kept it optional with undef as the default as this is truly an optional param (as are the other hashes). This will also ensure that it won't (at least the header/description) be rendered into the template when it's not supplied.

I've adjusted the data type to `Optional[Hash[String[1], Scalar]]`.

PR #380, opened 2021-05-07
Add system hooks management

carlosduelo:
A file hook will run on each event so it's up to you to filter events
or projects within a file hook code. You can have as many file hooks
as you want. Each file hook will be triggered by GitLab asynchronously
in case of an event. For a list of events see the system hooks
documentation.
Reference: http://10.191.3.4/help/administration/file_hooks


manifests/init.pp
```diff
@@ -115,6 +115,7 @@
   Stdlib::Absolutepath              $config_file                     = '/etc/gitlab/gitlab.rb',
   Optional[Hash]                    $consul                          = undef,
   Optional[String]                  $custom_hooks_dir                = undef,
+  Optional[String]                  $system_hooks_dir                = undef,
```
bastelfreak, line 118: should we only limit it to absolute paths or does something else makes sense?
```suggestion
  Optional[Stdlib::Absolutepath]    $system_hooks_dir                = undef,
```

manifests/init.pp
```diff
@@ -190,6 +191,7 @@
   Array                             $backup_cron_skips               = [],
   Hash                              $custom_hooks                    = {},
   Hash                              $global_hooks                    = {},
+  Hash                              $system_hooks                    = {},
```
bastelfreak, line 194: can we make this a bit stricter?
```suggestion
  Hash[String[1],Hash[String[1],String[,]] $system_hooks             = {},
```

manifests/system_hook.pp
```diff
@@ -0,0 +1,55 @@
+# @summary A file hook will run on each event so it's up to you to filter events or projects
```
bastelfreak, line 1: nice. good docs! :+1:

manifests/system_hook.pp
```diff
@@ -0,0 +1,55 @@
...
+  Stdlib::Absolutepath                          $system_hooks_dir,
+  Optional[String[1]]                           $content          = undef,
+  Optional[Pattern[/^puppet:/]]                 $source           = undef,
+) {
+  if $system_hooks_dir {
+    $hook_path = $system_hooks_dir
+  } elsif $gitlab::system_hooks_dir {
+    $hook_path = $gitlab::system_hooks_dir
+  } else {
+    $hook_path = '/opt/gitlab/embedded/service/gitlab-rails/file_hooks'
```
bastelfreak, line 26: can't we set that as default in the init.pp? that would make the logic in this defined resource way shorter?

spec/defines/system_hook_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  context "with type => #{type} and source" do
+    let(:source) { 'puppet:///modules/my_module/file-hook' }
+    let(:params) do
+      {
+        custom_hooks_dir: '/custom/hooks/dir',
+        source: source
+      }
+    end
+
+    it { is_expected.to compile }
```
bastelfreak, line 23: please ensure that the whole catalog compiles:
```suggestion
    it { is_expected.to compile.with_all_deps }
```

bastelfreak:
thanks for the PR! I made some tiny inline comments, can you please take a look?

manifests/system_hook.pp
carlosduelo: sure!!

carlosduelo:
@bastelfreak for your help! I think I resolved all the your comments

manifests/init.pp
```diff
@@ -91,105 +91,107 @@
...
-  Hash                              $custom_hooks                    = {},
-  Hash                              $global_hooks                    = {},
+  String                              $rake_exec                       = '/usr/bin/gitlab-rake',
+  Optional[Hash]                      $alertmanager                    = undef,
+  Optional[Hash]                      $ci_redis                        = undef,
+  Optional[Hash]                      $ci_unicorn                      = undef,
+  Boolean                             $config_manage                   = true,
+  Stdlib::Absolutepath                $config_file                     = '/etc/gitlab/gitlab.rb',
+  Optional[Hash]                      $consul                          = undef,
+  Optional[String]                    $custom_hooks_dir                = '/opt/gitlab/embedded/service/gitlab-shell/hooks',
```
bastelfreak, line 117: since those two variables now have a default value, they cannot be Optional (because you cannot pass undef to it).
when when we already update it we can update the datatype as well
```suggestion
  Stdlib::Absolutepath                $custom_hooks_dir                = '/opt/gitlab/embedded/service/gitlab-shell/hooks',
```

manifests/init.pp
```diff
@@ -91,105 +91,107 @@
...
+  Optional[Stdlib::Absolutepath]      $system_hooks_dir                = '/opt/gitlab/embedded/service/gitlab-rails/file_hooks',
```
bastelfreak, line 118:
```suggestion
  Stdlib::Absolutepath   $system_hooks_dir                = '/opt/gitlab/embedded/service/gitlab-rails/file_hooks',
```
aligning the `=` is quite impossible in this code suggestion :D

bastelfreak:
thanks for all the work!

PR #403, opened 2022-10-05
Add option to disable output of config diff

### LINKED ISSUE #363: The GitLab config file resource should not show diffs

- Puppet: All
- Ruby: All
- Distribution: All
- Module version: All

https://github.com/voxpupuli/puppet-gitlab/blob/master/manifests/omnibus_config.pp#L110-L121 should have `show_diff` set to `false`.

Given the discussion in #301, users can no longer set the root password after gitlab has been initially configured. Given this, the root password must be passed into the configuration file directly and will be output to the resource diff when updated which leaks the password into the logs.

Alternatively, the entire content string could be marked as `Sensitive`.

The ability to set the [root password securely](https://github.com/voxpupuli/puppet-gitlab/issues/301#issuecomment-487211322) would be preferred. 

Sensitive information should not be leaked into the run logs.

teluq-pbrideau:
I configure my authentication through `azure_activedirectory_v2`, and I don’t want my `client_secret` to be displayed in clear in the puppet logs.

Converting the `gitlab.rb.erb` to `.epp` so it supports the `sensitive` type would probably be cleaner, but I did not have success with it.  So this PR add option to disable output of the diff for sensitive information

```puppet
class example {
  class { 'gitlab' :
    [...]
    config_show_diff => false,
  }
}
```

I’ve included the option to also affect the file containing `$pgbouncer_password`, as it is sensitive information.  Is it is already a `epp` template, it could support `sensitive` type, but as I did not convert to the `sensitive` type elsewhere, it seemed simpler this way.

Feel free to comment if I missed anything.

Fixes #363

### DIFF
--- a/manifests/host_config.pp
+++ b/manifests/host_config.pp
@@ -7,6 +7,7 @@
 # @param pgpass_file_ensure Create .pgpass file for pgbouncer authentication. When set to present requires valid value for pgbouncer_password.
 # @param pgpass_file_location Path to location of .pgpass file used by consul to authenticate with pgbouncer database
 # @param pgbouncer_password Password for the gitlab-consul database user in the pgbouncer database
+# @param config_show_diff Whether to display diff in the puppet log or not
 class gitlab::host_config (
   $config_dir = '/etc/gitlab',
   $skip_auto_migrations = $gitlab::skip_auto_migrations,
--- a/manifests/host_config.pp
+++ b/manifests/host_config.pp
@@ -15,6 +15,7 @@ class gitlab::host_config (
   $pgpass_file_ensure = $gitlab::pgpass_file_ensure,
   $pgpass_file_location = $gitlab::pgpass_file_location,
   $pgbouncer_password = $gitlab::pgbouncer_password,
+  $config_show_diff = $gitlab::config_show_diff,
 ) {
   file { $config_dir:
     ensure => 'directory',
@@ -87,10 +88,11 @@ class gitlab::host_config (
   } else {
     # owner,group params for pgpass_file should NOT be changed, as they are hardcoded into gitlab HA db schema for pgbouncer database template
     file { $pgpass_file_location:
-      ensure  => $pgpass_file_ensure,
-      owner   => 'gitlab-consul',
-      group   => 'gitlab-consul',
-      content => epp('gitlab/.pgpass.epp', {
+      ensure    => $pgpass_file_ensure,
+      owner     => 'gitlab-consul',
+      group     => 'gitlab-consul',
+      show_diff => $config_show_diff,
+      content   => epp('gitlab/.pgpass.epp', {
           'pgbouncer_password' => $pgbouncer_password,
       }),
     }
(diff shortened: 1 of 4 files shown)


manifests/host_config.pp
```diff
@@ -15,6 +15,7 @@
   $pgpass_file_ensure = $gitlab::pgpass_file_ensure,
   $pgpass_file_location = $gitlab::pgpass_file_location,
   $pgbouncer_password = $gitlab::pgbouncer_password,
+  $config_show_diff = $gitlab::config_show_diff,
```
smortex, line 18: Documentation of this new class parameter should be added above.

manifests/omnibus_config.pp
```diff
@@ -4,7 +4,8 @@
 # @param config_file Path of the Gitlab Omnibus config file.
 class gitlab::omnibus_config (
   $config_manage = $gitlab::config_manage,
-  $config_file = $gitlab::config_file
+  $config_file = $gitlab::config_file,
+  $config_show_diff = $gitlab::config_show_diff,
```
smortex, line 8: Documentation of this new class parameter should be added above.

smortex (commented):
Thanks! I wonder if these parameters are required after all, and if we _really_ want this to be optional if it should rather default to `false`.

teluq-pbrideau:
Would you like me to add the `config_show_diff` option to `global_hook.pp`, `system_hook.pp` and `custom_hook.pp`, and set the default value to `false`?  It would give option to people who want to display the diff be able to do it.

teluq-pbrideau:
@smortex I’ve stumble over this today, is there anything else I could do to make the change to get merged?

root-expert:
@teluq-pbrideau Please rebase with out master branch and it can be merged afterwards

manifests/init.pp
```diff
@@ -114,6 +115,7 @@
   Optional[Hash]                      $ci_unicorn                      = undef,
   Boolean                             $config_manage                   = true,
   Stdlib::Absolutepath                $config_file                     = '/etc/gitlab/gitlab.rb',
+  Boolean                             $config_show_diff                = false,
```
smortex, line 118: As far as I can tell, the previous code did show the diff so if we want to be backward compatible, this should default to true?

Also, when we switch to epp templates, this setting would be useless if I recall correctly because in this case the diff is not shown when it contains sensitive data.

smortex (commented):
As requested by @root-expert, can you please rebase your changes on top of the main branch?

From your working directory:

```
git fetch origin         # Download the latest code we have here
git rebase origin/master # Move your commits on top of the main branch
git push -f              # Send the changes (-f is required because we re-wrote history)
```

manifests/init.pp
teluq-pbrideau: > As far as I can tell, the previous code did show the diff so if we want to be backward compatible, this should default to true?

That is also what I thought previously, and was then asked to put a secure default value.  But any value is OK for me, i’ll set the one you want!

> 
> Also, when we switch to epp templates, this setting would be useless if I recall correctly because in this case the diff is not shown when it contains sensitive data.

As said in the PR description, I first tried to convert to an `epp` template without success. As it was more than 3 month ago, I do not remember why exactly…  But you are right, once the template is converted to `epp`, the strings marked as `Sensitive()` will be obfuscated.

teluq-pbrideau:
Closing as it seems this is not wanted

PR #414, opened 2023-06-14
Relax dependencies version requirements

smortex:
* Allow puppetlabs-stdlib 9.x


zilchms:
i believe this was addressed in #415, do you think we can close this?

smortex:
Indeed, thanks @zilchms!

PR #421, opened 2023-10-06
Adding support for the gitlab_sshd configuration hash

### LINKED ISSUE #417: gitlab_sshd support in the omnibus config

- Puppet: 7.25
- Ruby: 2.7.8p225
- Distribution: Rocky 8.8
- Module version: 8.1.0

Hi,

We currently cannot configure the [gitlab_sshd part of the omnibus config](https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/files/gitlab-config-template/gitlab.rb.template#L1201).

Sample code:

```
      gitlab_sshd => {
         enable => true,
         listen_address => 'git.example.com:2222',
      },
```

Error seen in Puppet:

`Class[Gitlab]: has no parameter named 'gitlab_sshd'`

The documentation states

`Please note, managing the sshd service and openssh is outside the scope of this module.`

Does this fall under that umbrella?  Is there any other way to put these options into the config file?  If we add them manually or via other puppet code they are overwritten.

lodzo:
Adding support for the gitlab_sshd configuration

Fixes #417

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,12 +434,14 @@ initial connections for ssh operations. GitLab has created a feature that allows
 authorized ssh keys to be stored in the db (instead of the `authorized_keys`
 file for the `git` user)
 
-You can enable this feature in GitLab using the `store_git_keys_in_db` parameter.
+You can enable this feature in GitLab using the `store_git_keys_in_db` parameter,
+or by enabling `gitlab-sshd` as it is configured to use fast lookup automatically.
 
-Please note, managing the sshd service and openssh is outside the scope of this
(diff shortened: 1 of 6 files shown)


smortex (commented):
LGTM.  What do you think about updating the README since #417 says it is confusing?  At least we can remove the bit that says that the module cannot manage SSH, but if you have experience with this you may want to provide some guidance?

lodzo:
I'll try and reword that section of the README

smortex (approved):
Thanks for the head-up!

LGTM!

PR #423, opened 2023-11-23
Remove legacy top-scope syntax

smortex:


smortex:
Bogus entry, no more info and no context. Labelled skip-changelog and should not appear in future releases.

PR #468, opened 2025-07-06
Implement gitlab settings type & provider

jay7x:


lib/puppet_x/voxpupuli/gitlab/gitlab_settings.json
smortex, line 1: How was this produced?  I guess typing has to be handled manually, but the rest has been auto-generated. I wonder if we should bundle the code that merge new settings in this repo, and add some JSON schema to assert that a type was appropriately set?  We way also need support for a deprecated key in the future if settings are removed.

jay7x: Hi @smortex! This file is generated by the rake task (see the PR just after the json file). There are 2 tasks. First one fetches documentation and parses settings out of it to a temporary json file. Second generates the json file for helper based on the temporary json file. Also see the yaml file in next to the rake tasks. It allows to remove and to override the settings types and descriptions (and to add behaviour if needed)

PR #470, opened 2025-08-01
feat: implement initial root API token management

jay7x:
This PR implements somehow hack-ish, but the only possible way to create an initial root API token during the Gitlab installation. This allows to start managing Gitlab resources via API immediately after it is installed.

While I understand how bad it looks, I guess the benefits are big enough..


yakatz (changes requested):
I don't mind the "hack-ish" approach, but some other thoughts:
- The documentation should have a warning about how this could be insecure.
- A change like this needs an acceptance test to make sure the script actually works.
- I would also like to see a test that it doesn't run the script twice - it shouldn't the way it is written now, but it would be good to have for the future.

templates/create_initial_root_token.rb.epp
```diff
@@ -0,0 +1,21 @@
+<%-|
+  Optional[Sensitive[String[1]]] $token,
+  Integer[0] $token_ttl_minutes,
+  Stdlib::AbsolutePath $token_file_path,
+|-%>
+# This scripts creates an initial root token and stores it to the
```
TheMeier, line 6: how about a shebang line?

yakatz: It's executed through the Gitlab Rails Runner, so it shouldn't need one.

PR #480, opened 2025-12-04
workaround: lint check disabled. See description for details

marcusdots:
In my opinion the requirement of the linter is satisfied. The code is idempotent. It is the parameter passing using the splat operator that enables idempotency. Puppet lint is a static validator that does not see the runtime fullfilment. It is ok to disable the check


Joris29 (approved):

tuxmea:
Errors coming from installing gitlab-ce:
```
STDERR: sysctl: permission denied on key "kernel.pid_max"
  sysctl: permission denied on key "fs.protected_fifos"
  sysctl: permission denied on key "fs.protected_hardlinks"
  sysctl: permission denied on key "fs.protected_regular"
  sysctl: permission denied on key "fs.protected_symlinks"
```

tuxmea:
In this case the idempotency is coming from a hash which is used on the exec resource using splat operator.
The lint check can not evalute this.

Therefore it is OK to disable the idempotency check.

PR #487, opened 2026-02-16
[security] Sensitive type for gitlab_rails

m8t:

LDAP crendentials can be passed through gitlab_rails, which will appear in clear in Puppetdb for the class attributes.

This commit permits to hide from the class attributes inside PuppetDB, but still renders the password visible inside PuppetDB with the generated content of the template.

https://github.com/voxpupuli/puppet-gitlab/issues/486

### DIFF
--- a/templates/gitlab.rb.erb
+++ b/templates/gitlab.rb.erb
@@ -63,14 +63,21 @@ roles <%= decorate(@roles) %>
 git_data_dirs(<%= decorate(@git_data_dirs) %>)
 
 <%- end -%>
-<%- if @gitlab_rails -%>
+<%-
+if @gitlab_rails
+  if @gitlab_rails.is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
+    gitlab_rails = @gitlab_rails.unwrap
+  else
+    gitlab_rails = @gitlab_rails
+  end
+-%>
 
 ############################
 # gitlab.yml configuration #
 ############################
 
-<%- @gitlab_rails.keys.sort.each do |k| -%>
-gitlab_rails['<%= k -%>'] = <%= decorate(@gitlab_rails[k]) %>
+<%- gitlab_rails.keys.sort.each do |k| -%>
+gitlab_rails['<%= k -%>'] = <%= decorate(gitlab_rails[k]) %>
 <%- end end -%>
 <%- if @user -%>
 
--- a/templates/gitlab.rb.erb
+++ b/templates/gitlab.rb.erb
@@ -63,21 +63,14 @@ roles <%= decorate(@roles) %>
 git_data_dirs(<%= decorate(@git_data_dirs) %>)
 
 <%- end -%>
-<%-
-if @gitlab_rails
-  if @gitlab_rails.is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
-    gitlab_rails = @gitlab_rails.unwrap
-  else
-    gitlab_rails = @gitlab_rails
-  end
--%>
+<%- if @gitlab_rails_unwrap -%>
 
 ############################
 # gitlab.yml configuration #
 ############################
 
-<%- gitlab_rails.keys.sort.each do |k| -%>
-gitlab_rails['<%= k -%>'] = <%= decorate(gitlab_rails[k]) %>
+<%- @gitlab_rails_unwrap.keys.sort.each do |k| -%>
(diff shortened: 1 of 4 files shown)


TheMeier:
```
REFERENCE.md is outdated; to regenerate: bundle exec rake strings:generate:reference
```

m8t:
> ```
> REFERENCE.md is outdated; to regenerate: bundle exec rake strings:generate:reference
> ```

Indeed, updated!

templates/gitlab.rb.erb
```diff
@@ -63,14 +63,21 @@ roles <%= decorate(@roles) %>
 git_data_dirs(<%= decorate(@git_data_dirs) %>)
 
 <%- end -%>
-<%- if @gitlab_rails -%>
+<%-
+if @gitlab_rails
+  if @gitlab_rails.is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
```
alexjfisher: Is this conditional needed?  I thought perhaps for some time, `unwrap` was a noop if the data wasn't Sensitive?

alexjfisher:
```
puppet apply -t -e 'notice("foo".unwrap)'
Notice: Scope(Class[main]): foo
```

m8t: On a type String it bailed out with an error, none-existing method

**edit** fails with type Hash, String is fine in ERB as long as you call it through scope.call_function

m8t: Thing is, in a ERB template it fails, in Puppet code it's fine, I think.

m8t: Tested and fails both by calling Puppet function or plain ERB:

```
gitlab_rails = scope.call_function('unwrap', @gitlab_rails)

  Filepath: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/pops/functions/dispatcher.rb
  Line: 40
  Detail: 'unwrap' expects 1 argument, got 16
```

```
gitlab_rails = @gitlab_rails.unwrap

  Filepath: /etc/puppetlabs/code/environments/production/modules/gitlab/templates/gitlab.rb.erb
  Line: 68
  Detail: undefined method `unwrap' for #<Hash:0x37327d95>
```

alexjfisher: In `erb` you generally have to use `scope.call_function` to call functions.  ie This form.

```
scope.call_function('unwrap',['foo'])
```

alexjfisher: > Tested and fails both by calling Puppet function or plain ERB:
> 
> ```
> gitlab_rails = scope.call_function('unwrap', @gitlab_rails)
> ```

Close!  But the second argument needs to be an array of arguments to the function.
ie `scope.call_function('unwrap', [@gitlab_rails])` should work I believe.

m8t: Oddly this works fine `puppet apply -t -e 'notice({"foo"=>0,"bar"=>1}.unwrap)'`

m8t: Works fine through Puppet code only to unwrap a type Hash.

```diff

diff --git a/manifests/omnibus_config.pp b/manifests/omnibus_config.pp
index bcef084..1720747 100644
--- a/manifests/omnibus_config.pp
+++ b/manifests/omnibus_config.pp
@@ -27,3 +27,3 @@ class gitlab::omnibus_config (
   $gitlab_pages = $gitlab::gitlab_pages
-  $gitlab_rails = $gitlab::gitlab_rails
+  $_gitlab_rails = $gitlab::gitlab_rails
   $gitlab_sshd = $gitlab::gitlab_sshd
@@ -121,2 +121,3 @@ class gitlab::omnibus_config (
     } else {
+      $gitlab_rails = $_gitlab_rails.unwrap
       file { $config_file:
```
works fine. I can send a new patch, avoids having type check in ERB

m8t: > Close! But the second argument needs to be an array of arguments to the function. ie `scope.call_function('unwrap', [@gitlab_rails])` should work I believe.

Ah true, thought about it, and forgot, I prefer this variant. I already sent a different commit, where unwrap is called from the manifest, reused in the template. In the end both are fine, and the odd condition around type checking is not needed.

alexjfisher (approved):

templates/gitlab.rb.erb
alexjfisher: Either works for me.  I've approved it as-is, but if you wanted to change it, I'm happy to reapprove.

